PR #9, opened 2023-07-18
[Feature 0] Basic Entities

### LINKED ISSUE #8: [Feature 0] Basic Entities
A feature containing the basic entity classes, including Category, Expense, MonthlyStorage, and SessionStorage. This implementation of the base entities will be used as a base to build the rest of our project. For this reason, the implementation is rough and does not contain all necessary Javadocs, and needs better testing. I will create a new Issue to address these concerns.

katarinaavucic:
Added base entities and tests as described in CRC cards. Also added an exception for any error with the entities (that may be used later). Please read all methods and documentation so that we can ensure they are good before building the rest of our project!


src/main/java/entities/MonthlyStorage.java
```diff
@@ -0,0 +1,50 @@
...
+        this.currentMonth = currentMonth;
+        this.monthlyBudget = monthlyBudget;
+
+        Category other = new Category("Other", 0);
+        this.categoryData = new ArrayList<>();
+        this.categoryData.add(other);
+        this.expenseData = new ArrayList<>();
+    }
+
+    public void setCategoryData(Category category){
```
aricasas: Could this be named addCategory instead? I think that would be more accurate to what the method does

src/main/java/entities/MonthlyStorage.java
```diff
@@ -0,0 +1,50 @@
...
+        this.categoryData.add(category);
+    }
+
+    public void setExpenseData(Expense expense){
```
aricasas: Also in here maybe change name to addExpense

src/main/java/entities/SessionStorage.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Creates a new SessionStorage and assigns empty values to both its instance attributes
+     */
+    public SessionStorage(){
+        this.monthlyData = new ArrayList<>();
+        this.recurData = new ArrayList<>();
+    }
+
+    public void setMonthlyData(MonthlyStorage month){
```
aricasas: can we change the name of this one to addMonthlyData? Cause this one is adding something to the array instead of setting it

src/main/java/entities/SessionStorage.java
```diff
@@ -0,0 +1,38 @@
...
+        this.monthlyData.add(month);
+    }
+
+    public void setRecurData(Expense recurExpense){
```
aricasas: change to addRecurExpense?

src/main/java/entities/SessionStorage.java
```diff
@@ -0,0 +1,38 @@
...
+        this.recurData.add(recurExpense);
+    }
+
+    public MonthlyStorage getMonthlyData(int currentMonth) throws EntityException {
```
aricasas: Could we change the parameter name to just month? Because this method should work with any month right, not just the current one

aricasas (changes requested):
This seems great! I just added some comments about things I think could be renamed, but other than that it all looks good to me. Thanks Kat!

luluwuwu (dismissed):
Everything looks great to me!!

src/main/java/entities/Category.java
YinLingithub, line 1: 1. Don't need colon after "@param _parameter_". 
2. Javadoc are needed above the classes too. 
These points applies to all files.

YinLingithub (changes requested):
1. Don't need colon after "@param _parameter_". 
2. Javadoc are needed above the classes too. 
These points applies to all files.

src/main/java/entities/MonthlyStorage.java
YinLingithub, line 1: Also, we need setters to edit expenses and categories.

aricasas (approved):

luluwuwu (approved):
Remember to complete javadoc!

YinLingithub (approved):

PR #11, opened 2023-07-28
[Feature 1] Load a Session

### LINKED ISSUE #1: [Feature 1] Load a session
A menu to create a new session or load a session from a file and open the main menus.

aricasas:
Implemented the Load Session Menu with all its functionality. Also created an empty Main Menu that opens up after loading a session. 

Functions are documented with JavaDoc and there are tests for my UCI and the gateway that works with file loading and saving.

I had to add a couple methods to some entities in order to implement all of my functionality but they shouldn't affect anything that we already had.

I also created a Main class where the VIews are built and the program is run.

Please check my code and feel free to ask any questions and make comments about things that I might have missed and/or could improve.

### DIFF
--- a/src/main/java/Main.java
+++ b/src/main/java/Main.java
@@ -1,0 +1,39 @@
+import entities.SessionStorage;
+import use_cases.session_load.SessionLoadUCI;
+import views.file_session_storage.FileSessionStorage;
+import views.main_menu.MainMenuV;
+import views.session_load.SessionLoadC;
+import views.session_load.SessionLoadP;
+import views.session_load.SessionLoadMenuV;
+
+import javax.swing.*;
+import java.awt.*;
+
+public class Main {
+    public static void main(String[] args) {
+        // Build the main program window
+        JFrame application = new JFrame("Expenditure Tracking");
+        application.setDefaultCloseOperation(WindowConstants.EXIT_ON_CLOSE);
+        CardLayout cardLayout = new CardLayout();
+        JPanel screens = new JPanel(cardLayout);
+        application.add(screens);
+
+        // Program session
+        SessionStorage session = new SessionStorage();
+
+        // Building menu views
+        SessionLoadMenuV sessionLoadMenu = new SessionLoadMenuV();
+        MainMenuV mainMenuV = new MainMenuV();
+
+        // Create and set controllers
+        sessionLoadMenu.setController(new SessionLoadC(new SessionLoadUCI(new FileSessionStorage(), new SessionLoadP(sessionLoadMenu, mainMenuV), session)));
+
+        // Add screens
+        screens.add(sessionLoadMenu);
+        screens.add(mainMenuV);
+
+        // Open program UI
+        application.pack();
+        application.setVisible(true);
+    }
+}
--- a/src/main/java/Main.java
+++ b/src/main/java/Main.java
@@ -13,24 +13,28 @@ public class Main {
     public static void main(String[] args) {
         // Build the main program window
         JFrame application = new JFrame("Expenditure Tracking");
+        application.setMinimumSize(new Dimension(500, 300));
         application.setDefaultCloseOperation(WindowConstants.EXIT_ON_CLOSE);
         CardLayout cardLayout = new CardLayout();
+
         JPanel screens = new JPanel(cardLayout);
         application.add(screens);
 
         // Program session
         SessionStorage session = new SessionStorage();
 
-        // Building menu views
-        SessionLoadMenuV sessionLoadMenu = new SessionLoadMenuV();
-        MainMenuV mainMenuV = new MainMenuV();
+        // Building menus
+        // Main Menu
+        MainMenuV mainMenu = new MainMenuV();
 
-        // Create and set controllers
-        sessionLoadMenu.setController(new SessionLoadC(new SessionLoadUCI(new FileSessionStorage(), new SessionLoadP(sessionLoadMenu, mainMenuV), session)));
+        // Load Session Menu
+        SessionLoadUCI sessionLoadUCI = new SessionLoadUCI(new FileSessionStorage(), new SessionLoadP(), session);
+        SessionLoadC sessionLoadController = new SessionLoadC(sessionLoadUCI);
+        SessionLoadMenuV sessionLoadMenu = new SessionLoadMenuV(sessionLoadController, mainMenu);
 
         // Add screens
         screens.add(sessionLoadMenu);
-        screens.add(mainMenuV);
+        screens.add(mainMenu);
 
         // Open program UI
         application.pack();
--- a/src/main/java/entities/SessionStorage.java
+++ b/src/main/java/entities/SessionStorage.java
@@ -1,12 +1,13 @@
 package entities;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Objects;
 
 /**
  *
  */
-public class SessionStorage {
+public class SessionStorage implements Serializable {
 
     private final ArrayList<MonthlyStorage> monthlyData;
     private final ArrayList<Expense> recurData;
--- a/src/main/java/entities/SessionStorage.java
+++ b/src/main/java/entities/SessionStorage.java
@@ -86,4 +86,26 @@ public class SessionStorage implements Serializable {
         this.recurData.clear();
         this.recurData.addAll(other.recurData);
     }
+
+    /**
+     * Check if this SessionStorage is equal to Object
+     * @param obj Any instance of Object
+     * @return True if equals, False is not equals
+     */
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (this.getClass() != obj.getClass()){
+            return false;
+        }
+
+        SessionStorage other = (SessionStorage) obj;
+        return Objects.equals(this.getRecurData(), other.getRecurData())
+                && Objects.equals(this.getAllMonthlyData(), other.getAllMonthlyData());
+    }
 }
--- a/src/main/java/entities/SessionStorage.java
+++ b/src/main/java/entities/SessionStorage.java
@@ -66,4 +66,16 @@ public class SessionStorage implements Serializable {
     public ArrayList<Expense> getRecurData(){
         return this.recurData;
     }
+
+    /**
+     * Copies all the contents from another SessionStorage object into this one.
+     * This method can be used to mutate the entire contents of a SessionStorage object in-place.
+     * @param other the SessionStorage object to copy the data from
+     */
+    public void copyDataFrom(SessionStorage other) {
+        this.monthlyData.clear();
+        this.monthlyData.addAll(other.monthlyData);
+        this.recurData.clear();
+        this.recurData.addAll(other.recurData);
+    }
 }
--- a/src/main/java/entities/SessionStorage.java
+++ b/src/main/java/entities/SessionStorage.java
@@ -60,6 +60,14 @@ public class SessionStorage implements Serializable {
     }
 
     /**
+     * Returns the list of the MonthlyData objects stored in this Session
+     * @return list of all MonthlyData objects
+     */
+    public ArrayList<MonthlyStorage> getAllMonthlyData() {
+        return this.monthlyData;
+    }
+
+    /**
      * Returns the list of recurring expenses in this Session
      * @return list of recurring expenses
      */
--- a/src/main/java/use_cases/session_load/SessionLoadException.java
+++ b/src/main/java/use_cases/session_load/SessionLoadException.java
@@ -2,7 +2,7 @@ package use_cases.session_load;
 
 public class SessionLoadException extends Exception {
     /**
-     * Creates a new SessionLoadException for reporting exceptions specific to the this use case.
+     * An exception used when there was an issue loading a session from a file
      */
     public SessionLoadException(String message) {
         super(message);
--- a/src/main/java/use_cases/session_load/SessionLoadException.java
+++ b/src/main/java/use_cases/session_load/SessionLoadException.java
@@ -1,0 +1,10 @@
+package use_cases.session_load;
+
+public class SessionLoadException extends Exception {
+    /**
+     * Creates a new SessionLoadException for reporting exceptions specific to the this use case.
+     */
+    public SessionLoadException(String message) {
+        super(message);
+    }
+}
--- a/src/main/java/use_cases/session_load/SessionLoadException.java
+++ b/src/main/java/use_cases/session_load/SessionLoadException.java
@@ -1,5 +1,8 @@
 package use_cases.session_load;
 
+/**
+ * An exception used when there was an issue loading a session from a file
+ */
 public class SessionLoadException extends Exception {
     /**
      * An exception used when there was an issue loading a session from a file
--- a/src/main/java/use_cases/session_load/SessionLoadIB.java
+++ b/src/main/java/use_cases/session_load/SessionLoadIB.java
@@ -1,5 +1,12 @@
 package use_cases.session_load;
 
 public interface SessionLoadIB {
+    /**
+     * Either loads a SessionStorage from a file, or creates a new empty one.
+     *
+     * @param inputData the user input data
+     * @return a SessionLoadOD object with the output data to be used by the View
+     * @throws SessionLoadException if the user chose to load a file and there is an error while opening it
+     */
     SessionLoadOD load(SessionLoadID inputData) throws SessionLoadException;
 }
--- a/src/main/java/use_cases/session_load/SessionLoadIB.java
+++ b/src/main/java/use_cases/session_load/SessionLoadIB.java
@@ -1,0 +1,5 @@
+package use_cases.session_load;
+
+public interface SessionLoadIB {
+    void load(SessionLoadID inputData);
+}
--- a/src/main/java/use_cases/session_load/SessionLoadIB.java
+++ b/src/main/java/use_cases/session_load/SessionLoadIB.java
@@ -1,5 +1,5 @@
 package use_cases.session_load;
 
 public interface SessionLoadIB {
-    void load(SessionLoadID inputData);
+    SessionLoadOD load(SessionLoadID inputData) throws SessionLoadException;
 }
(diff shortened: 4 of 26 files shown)


src/main/java/views/session_load/SessionLoadMenuV.java
```diff
@@ -0,0 +1,84 @@
...
+                if (j.showOpenDialog(null) == JFileChooser.APPROVE_OPTION) {
+                    try {
+                        outputData = this.controller.loadFile(j.getSelectedFile().getAbsolutePath());
+                        this.close();
+                        this.mainMenuV.openMainMenu(outputData.getMessage(), outputData.getSession());
+                    } catch (SessionLoadException e) {
+                        this.displayError(e.getMessage());
+                    }
+                }
+                break;
```
katarinaavucic, line 88: Just curious on what the "break" is for?

src/main/java/use_cases/session_load/SessionLoadIB.java
```diff
@@ -0,0 +1,5 @@
+package use_cases.session_load;
+
+public interface SessionLoadIB {
```
katarinaavucic, line 8: Missing documentation here :)

src/main/java/use_cases/session_load/SessionLoadUCI.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public SessionLoadOD load(SessionLoadID inputData) throws SessionLoadException {
+        if (inputData.getIsNew()) {
+            // Create a new empty SessionStorage
+            SessionStorage new_session = new SessionStorage();
+            this.sessionToModify.copyDataFrom(new_session);
+            return this.presenter.displaySuccess("Created new session successfully", this.sessionToModify);
+        } else {
+            // Load a SessionStorage from a file
+            try {
```
katarinaavucic, line 41: Why did you decide to only handle some errors and throw SessionLoadException? I am also a little confused about why you throw RuntimeException in the Controller after catching SessionLoadException.

src/main/java/views/session_load/SessionLoadC.java
```diff
@@ -0,0 +1,41 @@
...
+     *
+     * @return a SessionLoadOD object with the output data to be used by the View
+     */
+    public SessionLoadOD loadNew() {
+        SessionLoadID inputData = new SessionLoadID();
+        try {
+            return this.interactor.load(inputData);
+        } catch (SessionLoadException e) {
+            // Loading a new empty session should never fail
+            throw new RuntimeException();
```
katarinaavucic: I think I mentioned it above but I am a little confused on why you throw RuntimeException here instead of handling it and printing a message

katarinaavucic (commented):
Looks good!! I just had a few questions regarding bits of the code, but other than that things looked good! The only thing was that a lot of your interfaces and tests did not have Javadocs and same with some of the methods in some classes! Also in tutorial the TA said that we need Javadocs above each class, which were not in these files (honestly I forgot about it too until now). But everything else looks great!

src/main/java/use_cases/session_load/SessionLoadException.java
YinLingithub, line 1: Instead of saying "this use case" in the Javadoc, specify when and what exception occurs.

src/main/java/entities/SessionStorage.java
```diff
@@ -58,11 +59,53 @@ public MonthlyStorage getMonthlyData(int monthID) throws EntityException {
         throw new EntityException("That is not a valid currentMonth for this SessionStorage.");
     }
 
+    /**
+     * Returns the list of the MonthlyData objects stored in this Session
+     * @return list of all MonthlyData objects
```
YinLingithub, line 64: I think this part of the Javadoc should be written like "@return list ArrayList of all MonthlyData objects".

Make sure other Javadocs are also in the right format.

src/main/java/Main.java
```diff
@@ -0,0 +1,43 @@
...
+
+import javax.swing.*;
+import java.awt.*;
+
+public class Main {
+    public static void main(String[] args) {
+        // Build the main program window
+        JFrame application = new JFrame("Expenditure Tracking");
+        application.setMinimumSize(new Dimension(500, 300));
+        application.setDefaultCloseOperation(WindowConstants.EXIT_ON_CLOSE);
```
YinLingithub, line 17: I'm not sure if this line is not indented or it's because I'm not reviewing the code in Intellij? Make sure that all lines are indented correctly.

src/main/java/use_cases/session_load/SessionLoadOD.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Gets the session data that was loaded
+     *
+     * @return a SessionStorage object with the loaded data
+     */
+    public SessionStorage getSession() {
+        return session;
+    }
+}
```
YinLingithub, line 36: Follow Intellij suggestions to add setters. Although we might not need it, it is just a convention to have getters and setters that follows the Open/Closed Principle.

src/main/java/views/session_load/SessionLoadMenuV.java
```diff
@@ -0,0 +1,84 @@
...
+        JOptionPane.showMessageDialog(this, message);
+    }
+
+    /**
+     * Closes this menu by setting its visibility to false
+     */
+    @Override
+    public void close() {
+        this.setVisible(false);
+    }
```
YinLingithub, line 65: I'm not sure why do you need this method. It's just one line either way.

src/main/java/views/session_load/SessionLoadMenuV.java
```diff
@@ -0,0 +1,84 @@
...
+
+        JLabel title = new JLabel("Load session");
+        title.setAlignmentX(Component.CENTER_ALIGNMENT);
+
+        JPanel buttons = new JPanel();
+        JButton load_new = new JButton("Create new session");
+        JButton load_file = new JButton("Load session from a file");
+
+        load_new.addActionListener(this);
+        load_file.addActionListener(this);
```
YinLingithub, line 35: Although you used ActionListener, you did not implement the actionPerformed method from ActionListener interface.

YinLingithub (changes requested):
As indicated by comments.

src/main/java/use_cases/session_load/SessionLoadOD.java
katarinaavucic: Are you sure? I don't really see the need in adding methods that we aren't gonna use, it would just flag warnings in Intellij. Was this something the prof or TA said?

YinLingithub: This is just a convention that code writers do in real life. This is similar to the "Write List<Expenses> = new ArrayList... instead of ArrayList<Expenses> = new ArrayList..." that I mentioned in the group chat. Its not wrong as the way it is, but using List is more professional as it follows Open/Closed Principle. Similarly, it's not wrong to not have the setters, just that conventionally, you also create the setters so that any people who want to use our code can access the Entities without changing our code.

src/main/java/Main.java
aricasas: I think that might just be a display error? Because it does look indented to me on both IntelliJ and GitHub

src/main/java/use_cases/session_load/SessionLoadException.java
aricasas: Yeah, I just copied over what EntityException had and forgot to update the documentation better. Thanks for pointing this out. I just modified it

src/main/java/use_cases/session_load/SessionLoadUCI.java
aricasas: Oh, if what we're doing is just creating a new empty session that is just creating an object, so there's no exception to handle there. For the case of loading it from a file, I handle the IO and ClassNotFound exceptions here because those are the ones thrown by the gateway, and then I turn them into the SessionLoadException because that is the one specific to this use case and that the View can handle

src/main/java/use_cases/session_load/SessionLoadOD.java
aricasas: Well in this case the output data should just be modified by the presenter when creating it no? So adding public setters would just add the ability to modify the already created output data when it shouldn't actually be modified after its creation, so I kinda disagree on adding the setters. I'll wait for everyone to comment here so that we can decide and I can add them pretty quick if we need them.

src/main/java/views/session_load/SessionLoadMenuV.java
aricasas: It is implemented further below

src/main/java/use_cases/session_load/SessionLoadIB.java
aricasas: Yeah, for some reason I was in the idea that we only needed JavaDocs for the concrete implementations and not the interfaces themselves, but I corrected it and added documentation to all of them now

src/main/java/views/session_load/SessionLoadMenuV.java
aricasas: Oh, the break is just something that you need to add in switch statements in java. I think in python the equivalent would be a match statement and in those the break is not explicit, but in java at the end of every case you need to write that.

https://www.w3schools.com/java/java_switch.asp

src/main/java/views/session_load/SessionLoadMenuV.java
aricasas: Oh, cause its part of the output boundary, I guess I could write a method called displaySuccess in the output boundary that does both this and open the main menu, but I kinda like the readability this provides in the actionPerformed method below where you can see exactly what it is doing by first closing this menu and then opening it, because I feel like this.close() + this.mainMenuV.openMainMenu() is more descriptive even if its a bit more verbose

aricasas: If you feel strongly towards changing it I can change it no problem though

src/main/java/views/session_load/SessionLoadMenuV.java
YinLingithub: Yeah, sorry, I didn't see it.

src/main/java/views/session_load/SessionLoadC.java
aricasas: I didn't want to make the loadNew method throw SessionLoadException, because creating a new session is just creating a new SessionStorage object and that has no way to fail. But because the interactor uses the same load method for both an empty session (which never throws) or loading from a file (which may throw), the compiler forces me to handle the exception. 

So, the inside of the catch should never run at all no matter what, and if it ever *does* run, it means the program is in such a screwed up state that we somehow couldn't even create a SessionStorage object, and in that case the program should just crash because there's nothing else to do. Maybe changing it to an assertion instead of just throwing an exception could communicate the intent better?

src/main/java/views/session_load/SessionLoadMenuV.java
YinLingithub: Yeah, I'm still not sure how much difference it's going to make. I feel like it's just redundant code, and being more descriptive just made it harder to understand what it's doing, compared to this.setVisible(false) which clearly states that you are setting the window invisible.

src/main/java/views/session_load/SessionLoadMenuV.java
katarinaavucic: Ahh, I see. Thanks!

src/main/java/use_cases/session_load/SessionLoadIB.java
katarinaavucic: Sounds good! I think we will need to go over all documentation anyways so that everything looks coherent.

src/main/java/use_cases/session_load/SessionLoadUCI.java
katarinaavucic: Okay, makes sense thanks!

src/main/java/views/session_load/SessionLoadC.java
katarinaavucic: Hmmm I think what you said makes sense. Its up to you if you want to switch to an Assertion, but you're explanation makes sense to me. Maybe you could include a brief explanation in the method Javadoc?

src/main/java/use_cases/session_load/SessionLoadUCI.java
```diff
@@ -0,0 +1,48 @@
+package use_cases.session_load;
+
+import entities.SessionStorage;
+import views.file_session_storage.SessionStorageG;
```
katarinaavucic: I just went over Yin's and realized I forgot to check for package imports here so I am going back through yours now. This class should not know that the views package exists, so instead of creating a SessionStorageG object, you should create a Gateway Boundary similar to the Input/Output/View Boundaries that resides in the use_case package, and cast the Gateway to that GatewayBoundary when you are passing it in and storing it in the interaction!

src/main/java/views/file_session_storage/FileSessionStorage.java
```diff
@@ -0,0 +1,47 @@
+package views.file_session_storage;
+
+import entities.SessionStorage;
```
katarinaavucic, line 3: Only the use case layer (application business rules) should be accessing the entities package (Clean Architecture slides 04-page 12). We break that with the SessionStorage because there is no other way to access that information, so just make sure that for every class that you import it there is no other option (so you can explain it).

katarinaavucic: OB, OD, G, MainMenuVB, and P all have it imported.

katarinaavucic (commented):
Went over the package imports so I have a few more comments to make!

src/main/java/use_cases/session_load/SessionLoadOD.java
YinLingithub: I agree with you, Ari. It's just two different views where some think the code should be open to extension (especially so in the case of making a library); some think the code should be clean and minimal (especially when the same person will continue update the program). 

In our case, we are not making a library, and it is likely that only we will make changes to the code, so we do not need to write the setters. However, it will be our group design decision to include setters or not, so we must reach a consensus and make sure all use cases does the same.

src/main/java/views/session_load/SessionLoadMenuV.java
aricasas: Yeah after thinking about it a bit more I see that you were right now. I made a new displaySuccess method just like the displayError in the view boundary and implemented it so that it does both this.setVisible(false) and opens the main menu. The actionPerformed method is simpler now and it's clearer in displaySuccess what it does. Do you think this was a good solution?

src/main/java/use_cases/session_load/SessionLoadUCI.java
aricasas: Yeah, actually the SessionStorageG was supposed to be that boundary so that the interactor could communicate to the FIleSessionStorage which was the concrete class. But I accidentaly put the SessionStorageG in the worng package. I just moved it inside the use_cases.session_load package and that should be correct now

src/main/java/views/session_load/SessionLoadMenuV.java
YinLingithub: Yes, it looks good!

src/main/java/use_cases/session_load/SessionLoadOD.java
luluwuwu: As I mentioned in our gc, I will need setters for my input data to cast the user input of category budgets and expense amounts. (since we allow users to input any number type but we are handling them in terms of doubles.)

YinLingithub: Conclusion made from group chat: Only add setters in input/output data classes if they are required for the feature. 

I have removed the setters from my input/output data classes too.

src/main/java/views/file_session_storage/FileSessionStorage.java
katarinaavucic: Sorry, I meant only the G, MainMenuVB, and P since they are in the interface adapter layer! We could either make note of it in every Javadoc or state is elsewhere in our final submission? I am not sure we will likely need to discuss in group meeting.

aricasas: Okay. I'll wait for consensus before doing anything

src/test/java/views/file_session_storage/FileSessionStorageTest.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    @Test
+    public void FileSessionStorage() {
+        FileSessionStorage fileSessionStorage = new FileSessionStorage();
+
+        // Save session into file
+        try {
+            fileSessionStorage.save("FSSSaveFileTest.ser", sampleSession);
+        } catch (Exception e) {
+            Assertions.fail();
```
luluwuwu: It is a little unclear what Assertions.fail() is testing for. Maybe try to specify an exception? or explain in more detail in Javadoc. (this also includes the ones below)

src/main/java/views/file_session_storage/FileSessionStorage.java
katarinaavucic: Sounds good!

src/test/java/views/file_session_storage/FileSessionStorageTest.java
aricasas: Yes, I will rewrite the testing for FileSessionStorage to make it clearer, and also I needed to add tests for the fail cases

aricasas: Done

src/test/java/use_cases/session_load/SessionLoadUCITest.java
```diff
@@ -76,4 +76,54 @@ public void SessionLoadUCILoadFile() {
...
+
+        // Check that our session is left unchanged
+        Assertions.assertEquals(session, sampleSession);
+    }
+
+    /**
+     * Tests trying to load an invalid file  and that a SessionLoadException will be raised
+     */
+    @Test
+    public void SessionLoadLUCILoadInvalidFile() throws IOException {
```
YinLingithub: Just a typo here. Everything else in this file looks great.

luluwuwu (dismissed):
Everything looks good!

src/test/java/use_cases/session_load/SessionLoadUCITest.java
aricasas: Just corrected it

YinLingithub (approved):
Perfect!

katarinaavucic (approved):
Everything looks good!!

luluwuwu (approved):
All good!

PR #12, opened 2023-07-30
[Feature 4] Monthly Menu Completed

### LINKED ISSUE #6: [Feature 4] Monthly menu
A menu to connect all functionalities within the system: add/edit categories, add/edit expense, generate summary.

YinLingithub:
Implemented the Month Menu with its GUI, and initialized related ActionListener method. Related classes for updating the Month Menu are also implemented. 
All classes and methods are documented with Javadoc. 
There are tests for the use case interactor testing output for both success and fail cases.

### DIFF
--- a/src/main/java/use_cases/monthly_menu/MonthMenuOD.java
+++ b/src/main/java/use_cases/monthly_menu/MonthMenuOD.java
@@ -5,40 +5,80 @@ import entities.Expense;
 
 import java.util.ArrayList;
 
+/**
+ * The output data class for storing view output. The interactor
+ * class for getting the Month Menu output creates an object of
+ * this class, and pass it into the output boundary. The output
+ * boundary returns this type of object, and is then formatted
+ * and shown in the Month Menu view.
+ */
 public class MonthMenuOD {
     ArrayList<Expense> expenseData;
     ArrayList<Category> categoryData;
     String warning;
 
+    /**
+     * Constructs an MonthMenuOD holding output data.
+     * @param expenseData ArrayList of Expense stored in a MonthlyStorage
+     * @param categoryData ArrayList of Category stored in a MonthlyStorage
+     */
     public MonthMenuOD(ArrayList<Expense> expenseData, ArrayList<Category> categoryData){
         this.expenseData = expenseData;
         this.categoryData = categoryData;
     }
 
+    /**
+     * Constructs an MonthMenuOD holding error message.
+     * @param warning error message to show in the view
+     */
     public MonthMenuOD(String warning){
         this.warning = warning;
     }
 
+    /**
+     * Gets the expenseData stored.
+     * @return expenseData to be outputted
+     */
     public ArrayList<Expense> getExpenseData() {
         return expenseData;
     }
 
+    /**
+     * Gets the categoryData stored.
+     * @return categoryData to be outputted
+     */
     public ArrayList<Category> getCategoryData() {
         return categoryData;
     }
 
+    /**
+     * Gets the error message stored.
+     * @return warning message to be outputted
+     */
     public String getWarning() {
         return warning;
     }
 
+    /**
+     * Sets specified ArrayList of Expense to expenseData variable.
+     * @param expenseData ArrayList of Expense stored in a MonthlyStorage
+     */
     public void setExpenseData(ArrayList<Expense> expenseData) {
         this.expenseData = expenseData;
     }
 
+    /**
+     * Sets specified ArrayList of Category to categoryData variable.
+     * @param categoryData ArrayList of Category stored in a MonthlyStorage
+     */
     public void setCategoryData(ArrayList<Category> categoryData) {
         this.categoryData = categoryData;
     }
 
+    /**
+     * Sets specified String to warning variable.
+     * @param warning error message to show in the view
+     */
     public void setWarning(String warning) {
         this.warning = warning;
     }
--- a/src/main/java/use_cases/monthly_menu/MonthMenuOD.java
+++ b/src/main/java/use_cases/monthly_menu/MonthMenuOD.java
@@ -13,9 +13,9 @@ import java.util.ArrayList;
  * and shown in the Month Menu view.
  */
 public class MonthMenuOD {
-    ArrayList<Expense> expenseData;
-    ArrayList<Category> categoryData;
-    String warning;
+    private ArrayList<Expense> expenseData;
+    private ArrayList<Category> categoryData;
+    private String warning;
     boolean successful;
 
     /**
@@ -69,29 +69,5 @@ public class MonthMenuOD {
      */
     public boolean isSuccessful() {
         return successful;
-    }
-
-    /**
-     * Sets specified ArrayList of Expense to expenseData variable.
-     * @param expenseData ArrayList of Expense stored in a MonthlyStorage
-     */
-    public void setExpenseData(ArrayList<Expense> expenseData) {
-        this.expenseData = expenseData;
-    }
-
-    /**
-     * Sets specified ArrayList of Category to categoryData variable.
-     * @param categoryData ArrayList of Category stored in a MonthlyStorage
-     */
-    public void setCategoryData(ArrayList<Category> categoryData) {
-        this.categoryData = categoryData;
-    }
-
-    /**
-     * Sets specified String to warning variable.
-     * @param warning error message to show in the view
-     */
-    public void setWarning(String warning) {
-        this.warning = warning;
     }
 }
--- a/src/main/java/use_cases/monthly_menu/MonthMenuOD.java
+++ b/src/main/java/use_cases/monthly_menu/MonthMenuOD.java
@@ -1,0 +1,4 @@
+package use_cases.monthly_menu;
+
+public class MonthMenuOD {
+}
--- a/src/main/java/use_cases/monthly_menu/MonthMenuOD.java
+++ b/src/main/java/use_cases/monthly_menu/MonthMenuOD.java
@@ -1,27 +1,45 @@
 package use_cases.monthly_menu;
 
+import entities.Category;
+import entities.Expense;
+
+import java.util.ArrayList;
+
 public class MonthMenuOD {
-    Object[][] expenseList;
-    Object[][] categoryList;
+    ArrayList<Expense> expenseData;
+    ArrayList<Category> categoryData;
+    String warning;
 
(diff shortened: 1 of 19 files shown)


src/main/java/use_cases/monthly_menu/MonthMenuOD.java
```diff
@@ -0,0 +1,85 @@
+package use_cases.monthly_menu;
+
+import entities.Category;
```
katarinaavucic, line 3: Is the OutputData allowed to import the entity packages? Generally we should be using general Object types in the OutputData. I believe the SessionStorage in the InputData is such an exception since we have no other way of accessing the rest of the stuff without it, but these could be replaced (and probably should).

src/main/java/use_cases/monthly_menu/UpdateViewUCI.java
```diff
@@ -0,0 +1,42 @@
+package use_cases.monthly_menu;
+
+import entities.*;
+import views.monthly_menu.MonthMenuP;
```
katarinaavucic: This class should not know the views package exists. Instead, create the presenter prior to creating the interactor and pass it in upon initialization of the interactor (casting the presenter to type MonthMenuOB), so that the interactor only imports the MonthMenuOB class.

katarinaavucic: This is demonstrated in Main.java of UserLogin! :)

src/main/java/views/monthly_menu/MonthMenuV.java
```diff
@@ -0,0 +1,188 @@
+package views.monthly_menu;
+
+import entities.Category;
```
katarinaavucic, line 3: This class too should not be importing the Category and Expense classes. If you recall from our Clean Architecture Slides (04, slide 12), Controllers should not be importing Entities, so this definitely shouldn't be happening (unless you have a good reason for it and can explain!)

src/main/java/views/monthly_menu/MonthMenuV.java
```diff
@@ -0,0 +1,188 @@
...
+     * expenses/categories, and the button to generate summary. The Month
+     * Menu window also shows the list of expenses and categories currently
+     * stored in the MonthlyStorage. The fail view shows a pop-up window
+     * showing the error message.
+     * Code inspired by https://stackoverflow.com/questions/5621338/how-to-add-jtable-in-jpanel-with-null-layout
+     * @param controller the controller class to get output data
+     * @param session the SessionStorage holding the required MonthlyStorage
+     * @param monthID the monthID of the required MonthlyStorage
+     */
+    public MonthMenuV(views.monthly_menu.UpdateViewC controller, SessionStorage session, int monthID) {
```
katarinaavucic, line 41: I am thinking you might want to create smaller, private classes to delegate some of these tasks to, since this method is really long! for example, replacing anything that you had to comment an explanation for with a private method would work really well, especially since you'll already have half your Javadoc!

katarinaavucic (changes requested):
Overall, everything looked good! I did make a few comments on some specific parts that might be Clean Architecture violations and some code smells. I think we will also need to discuss a common format for Javadocs since ours are all different.

src/main/java/views/monthly_menu/MonthMenuV.java
YinLingithub: Unfortunately, this is the only problem I cannot solve! I've tried have more smaller private methods. However, considering that we need to add code for button reactions later on, I ended up having to put some of the methods back to the constructor.

src/main/java/use_cases/monthly_menu/UpdateViewUCI.java
YinLingithub: It's fixed!

src/main/java/views/monthly_menu/MonthMenuV.java
YinLingithub: Expense and Category are basic object types we have for conveying information. Thus if we remove the imports and change the outputdata to Object[][] type in the UpdateViewUCI to accomodate Java Swing, it is not adhering to SOLID in point that its not Open to importing other libraries.
 
Of course, this is a assignment, so we should consult the Prof and/or TA about this. I will be able to change this part any time, so for now, I think we should keep it like this.

src/main/java/use_cases/monthly_menu/MonthMenuOD.java
YinLingithub: Same as mentioned in MonthMenuV review.

src/main/java/views/monthly_menu/MonthMenuP.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public class MonthMenuP implements MonthMenuOB {
+
+    /**
+     * Pass in and return output data.
+     * @param output output to be shown in MonthMenuView
+     * @return MonthMenuOD object that contains output data
+     */
+    @Override
+    public MonthMenuOD createOutput(MonthMenuOD output) {
```
aricasas, line 20: Shouldn't the presenter be the one to create the output data object instead of receiving it and returning it immediately?

src/main/java/views/monthly_menu/MonthMenuV.java
```diff
@@ -0,0 +1,188 @@
...
+            categoryScroll.setBounds(15, 15, 333, 90);
+            rightLayout.add(categoryScroll, BorderLayout.CENTER);
+            rightLayout.setBorder(BorderFactory.createEmptyBorder(10, 10, 10, 10));
+
+            //Add both side to overall layout and set up frame
+            layout.add(leftLayout, BorderLayout.WEST);
+            layout.add(rightLayout, BorderLayout.CENTER);
+            frame.setContentPane(layout);
+        }
+        catch (Exception e){
```
aricasas: I think writing the specific Exception that is expected type here could be better

src/main/java/views/monthly_menu/MonthMenuV.java
aricasas: I think if you can't avoid it, a long constructor in the view should be ok, because most of the code is just UI stuff and it is pretty well commented

src/main/java/views/monthly_menu/MonthMenuP.java
YinLingithub: I referenced the UserLoginCleanArchitecture repository, and in there the use case interactor creates the output data object and presenter modifies it. In my case, there's nothing in particular that I need to add to the MonthMenuOD, thus I let the presenter directly returned it.

aricasas: Ahh okok, that makes sense

aricasas (commented):
I just left a couple of comments, and other than the stuff Kat pointed out everything seems pretty good to me. Very good JavaDocs!

src/main/java/views/monthly_menu/MonthMenuV.java
YinLingithub: Actually, as you point this out, I found that there may be some problem with this try catch. I'll get this figured out ASAP. Thanks a lot, Ari.

YinLingithub: The problem is fixed! As I confirmed again, there wasn't any exception for me to catch as I already caught in UpdateViewUCI, so I changed the try catch to if else.

src/main/java/views/monthly_menu/MonthMenuV.java
katarinaavucic: Sounds good!

src/main/java/views/monthly_menu/MonthMenuV.java
katarinaavucic: Makes sense!

katarinaavucic (dismissed):

src/main/java/use_cases/monthly_menu/MonthMenuOD.java
```diff
@@ -70,28 +70,4 @@ public String getWarning() {
     public boolean isSuccessful() {
         return successful;
     }
-
-    /**
-     * Sets specified ArrayList of Expense to expenseData variable.
-     * @param expenseData ArrayList of Expense stored in a MonthlyStorage
-     */
-    public void setExpenseData(ArrayList<Expense> expenseData) {
```
luluwuwu: I think I'll still need setters to update the list from adds and edits. Maybe I'm wrong?

src/main/java/use_cases/monthly_menu/MonthMenuOD.java
```diff
@@ -13,9 +13,9 @@
  * and shown in the Month Menu view.
  */
 public class MonthMenuOD {
```
luluwuwu, line 15: Do I get the arraylist of category names from here or is it from somewhere else?

src/main/java/use_cases/monthly_menu/MonthMenuOD.java
YinLingithub: You shouldn't be needing these setters, because firstly, they are not original list of Expense/Category stored in MonthlyStorage; secondly, you are probably adding/editing the MonthlyStorage in your use case interactor.

src/main/java/use_cases/monthly_menu/MonthMenuOD.java
YinLingithub: You get the arraylist of Category objects from the MonthMenuV, in the actionPerformed method, calling controller.getOutput(session, monthID).getCategoryData(). 

Actually, let me set session and monthID as instance variables in the constructor, so that they can be used from the actionPerformed method.

katarinaavucic (dismissed):
Looks good!

src/main/java/use_cases/monthly_menu/MonthMenuOD.java
luluwuwu: ok I see thanks

luluwuwu (dismissed):
As long as actionPerformed() is implemented, everything is perfection!

src/main/java/use_cases/monthly_menu/MonthMenuOD.java
luluwuwu: YES

katarinaavucic (approved):
Everything looks good!

luluwuwu (approved):
Perfection!

aricasas (approved):
Everything looks great!

PR #21, opened 2023-08-04
[Feature 3] Create a New Month Completed

### LINKED ISSUE #2: [Feature 3] Create a new month
A feature that initializes a new month for the user to record expenses.

YinLingithub:
Implemented the Creat New Month use case and call to Month Menu GUI. I will implement a call to openMainMenu method when Ari's branch get merged to main. 
All classes and methods are documented with Javadoc.
There are tests for the use case interactor testing output for both success and fail cases. The view is tested in local file using main method.

### DIFF
--- a/src/main/java/use_cases/create_new_month/NewMonthID.java
+++ b/src/main/java/use_cases/create_new_month/NewMonthID.java
@@ -13,7 +13,7 @@ public class NewMonthID {
     private final double budgetValue;
 
     /**
-     * Constructs an NewMonthID holding input data.
+     * Constructs a NewMonthID holding input data.
      * @param session the SessionStorage to store the new MonthlyStorage
      * @param monthID the monthID of the new MonthlyStorage
      * @param budgetValue the budget for the new MonthlyStorage
--- a/src/main/java/use_cases/create_new_month/NewMonthID.java
+++ b/src/main/java/use_cases/create_new_month/NewMonthID.java
@@ -10,7 +10,7 @@ import entities.SessionStorage;
 public class NewMonthID {
     private final SessionStorage session;
     private final int monthID;
-    private final int budgetValue;
+    private final double budgetValue;
 
     /**
      * Constructs an NewMonthID holding input data.
@@ -18,7 +18,7 @@ public class NewMonthID {
      * @param monthID the monthID of the new MonthlyStorage
      * @param budgetValue the budget for the new MonthlyStorage
      */
-    public NewMonthID(SessionStorage session, int monthID, int budgetValue) {
+    public NewMonthID(SessionStorage session, int monthID, double budgetValue) {
         this.session = session;
         this.monthID = monthID;
         this.budgetValue = budgetValue;
@@ -42,9 +42,10 @@ public class NewMonthID {
 
     /**
      * Gets the budget value stored.
+     *
      * @return double type budget for new MonthlyStorage to be created in interactor
      */
-    public int getBudgetValue() {
+    public double getBudgetValue() {
         return budgetValue;
     }
 }
--- a/src/main/java/use_cases/create_new_month/NewMonthID.java
+++ b/src/main/java/use_cases/create_new_month/NewMonthID.java
@@ -2,15 +2,21 @@ package use_cases.create_new_month;
 
 import entities.SessionStorage;
 
+/**
+ * The input data class for storing Create New Month use case inputs.
+ * The controller class of Create New Month use case creates an object
+ * of this class, and pass it to the interactor class.
+ */
 public class NewMonthID {
     private final SessionStorage session;
     private final int monthID;
     private final int budgetValue;
 
     /**
-     * @param session
-     * @param monthID
-     * @param budgetValue
+     * Constructs an NewMonthID holding input data.
+     * @param session the SessionStorage to store the new MonthlyStorage
+     * @param monthID the monthID of the new MonthlyStorage
+     * @param budgetValue the budget for the new MonthlyStorage
      */
     public NewMonthID(SessionStorage session, int monthID, int budgetValue) {
         this.session = session;
@@ -18,14 +24,26 @@ public class NewMonthID {
         this.budgetValue = budgetValue;
     }
 
+    /**
+     * Gets the session stored.
+     * @return SessionStorage accessed in interactor
+     */
     public SessionStorage getSession() {
         return session;
     }
 
+    /**
+     * Gets the monthID stored.
+     * @return int ID of new MonthlyStorage to be created in interactor
+     */
     public int getMonthID() {
         return monthID;
     }
 
+    /**
+     * Gets the budget value stored.
+     * @return double type budget for new MonthlyStorage to be created in interactor
+     */
     public int getBudgetValue() {
         return budgetValue;
     }
--- a/src/main/java/use_cases/create_new_month/NewMonthID.java
+++ b/src/main/java/use_cases/create_new_month/NewMonthID.java
@@ -1,0 +1,32 @@
+package use_cases.create_new_month;
+
+import entities.SessionStorage;
+
+public class NewMonthID {
+    private final SessionStorage session;
+    private final int monthID;
+    private final int budgetValue;
+
+    /**
+     * @param session
+     * @param monthID
+     * @param budgetValue
+     */
+    public NewMonthID(SessionStorage session, int monthID, int budgetValue) {
+        this.session = session;
+        this.monthID = monthID;
+        this.budgetValue = budgetValue;
+    }
+
+    public SessionStorage getSession() {
+        return session;
+    }
+
+    public int getMonthID() {
+        return monthID;
+    }
+
+    public int getBudgetValue() {
+        return budgetValue;
+    }
+}
--- a/src/main/java/use_cases/create_new_month/NewMonthOD.java
+++ b/src/main/java/use_cases/create_new_month/NewMonthOD.java
@@ -1,37 +1,45 @@
 package use_cases.create_new_month;
 
-import entities.SessionStorage;
-
+/**
+ * The output data class for storing view output. The interactor
+ * class for creating a new MonthlyStorage creates an object of
+ * this class, and pass it into the output boundary. The output
+ * boundary returns this type of object, and is then formatted
+ * and used in the Create New Month view.
+ */
 public class NewMonthOD {
-    SessionStorage session;
-    int monthID;
     String warning;
     boolean successful;
 
-
-    public NewMonthOD(SessionStorage session, int monthID, boolean successful) {
-        this.session = session;
-        this.monthID = monthID;
+    /**
+     * Constructs an MonthMenuOD holding output data.
+     * @param successful boolean indicating whether create new MonthlyStorage was successful
+     */
+    public NewMonthOD(boolean successful) {
         this.successful = successful;
     }
 
+    /**
+     * Constructs an NewMonthOD holding error message.
+     * @param warning error message to show in the view
+     * @param successful boolean indicating whether create new MonthlyStorage was successful
+     */
     public NewMonthOD(String warning, boolean successful){
         this.warning = warning;
         this.successful = successful;
     }
 
-    public SessionStorage getSession() {
-        return session;
-    }
-
-    public int getMonthID() {
-        return monthID;
-    }
-
+    /**
+     * Gets the error message stored.
+     * @return warning message to be outputted
+     */
     public String getWarning() {
         return warning;
     }
 
+    /**
+     * @return true if access to Expense/Category list was successful
+     */
     public boolean isSuccessful() {
         return successful;
     }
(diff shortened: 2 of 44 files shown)


src/main/java/use_cases/create_new_month/NewMonthID.java
```diff
@@ -0,0 +1,51 @@
...
+ * The controller class of Create New Month use case creates an object
+ * of this class, and pass it to the interactor class.
+ */
+public class NewMonthID {
+    private final SessionStorage session;
+    private final int monthID;
+    private final double budgetValue;
+
+    /**
+     * Constructs an NewMonthID holding input data.
```
katarinaavucic: Minor grammar error here (an instead of a).

src/main/java/use_cases/create_new_month/NewMonthOD.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * The output data class for storing view output. The interactor
+ * class for creating a new MonthlyStorage creates an object of
+ * this class, and pass it into the output boundary. The output
+ * boundary returns this type of object, and is then formatted
+ * and used in the Create New Month view.
+ */
+public class NewMonthOD {
+    String warning;
```
katarinaavucic: These should be "private final".

katarinaavucic: Or at least private.

src/main/java/use_cases/create_new_month/NewMonthOD.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Constructs an MonthMenuOD holding output data.
+     * @param successful boolean indicating whether create new MonthlyStorage was successful
+     */
+    public NewMonthOD(boolean successful) {
+        this.successful = successful;
+    }
+
+    /**
+     * Constructs an NewMonthOD holding error message.
```
katarinaavucic, line 23: Same grammar mistake here.

src/main/java/use_cases/create_new_month/NewMonthOD.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Gets the error message stored.
+     * @return warning message to be outputted
+     */
+    public String getWarning() {
+        return warning;
+    }
+
+    /**
+     * @return true if access to Expense/Category list was successful
```
katarinaavucic: This one is missing a general description.

src/main/java/use_cases/create_new_month/NewMonthUCI.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return NewMonthOD object that contains output data
+     */
+    @Override
+    public NewMonthOD createOutput(NewMonthID input) {
+        SessionStorage session = input.getSession();
+        int monthID = input.getMonthID();
+        double budgetValue = input.getBudgetValue();
+
+        // Check that no MonthlyStorage in this session has the same monthID
+        ArrayList<MonthlyStorage> allMonthArrayList = session.getAllMonthlyData();
```
katarinaavucic: I think this could be put into a private method to make the code shorter. That, or it could be implemented into SessionStorage as a method there since that would make sense as well.

src/main/java/views/create_new_month/NewMonthC.java
```diff
@@ -0,0 +1,36 @@
...
+import use_cases.create_new_month.NewMonthOD;
+
+/**
+ * The controller class for creating new MonthlyStorage. The view class
+ * calls this class to get the NewMonthOD object. This class generates the
+ * input data class and calls the input boundary interface to get the
+ * NewMonthOD object.
+ */
+public class NewMonthC {
+    final NewMonthIB inputBoundary;
```
katarinaavucic: This should be private.

src/main/java/views/create_new_month/NewMonthC.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Construct the controller class.
```
katarinaavucic, line 18: To keep up with your previous naming conventions you should specify NewMonthC instead of controller.

src/main/java/views/create_new_month/NewMonthC.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    /**
+     * Create and pass the NewMonthID object to the inputBoundary method call.
+     * @param session the SessionStorage to store the new MonthlyStorage
+     * @param monthID the monthID of the new MonthlyStorage
+     * @param budgetValue the budget for the new MonthlyStorage
+     * @return NewMonthOD object that contains output data
+     */
+    NewMonthOD getOutput(SessionStorage session, int monthID, double budgetValue) {
```
katarinaavucic, line 32: This method should be set as public.

src/main/java/views/create_new_month/NewMonthV.java
```diff
@@ -0,0 +1,151 @@
...
+import java.awt.event.ActionListener;
+import java.util.Objects;
+
+/**
+ * The view class for creating new MonthlyStorage. This class implements the
+ * ActionListener interface. This class calls the controller class to get the
+ * NewMonthOD object, and use the output to set up success and fail view.
+ */
+public class NewMonthV implements ActionListener, LoadMonthMenuVB {
+    NewMonthC controller;
```
katarinaavucic: Shouldn't these all be set to private too? Unless there is a reason they are default, I think we want to give them the most limited scopes if we can.

src/main/java/views/create_new_month/NewMonthV.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * Load Month Menu and notify user.
+     * @param session the SessionStorage holding the required MonthlyStorage
+     * @param monthID the monthID of the required MonthlyStorage
+     * @param message notify user that Month Menu is updated
+     */
+    @Override
+    public void loadMonthMenu(SessionStorage session, int monthID, String message) {
```
katarinaavucic: I can't recall if this was something said in CSC148 or just a preference, but I put all the private methods below the public ones. I can't recall if this was an actual thing or not, but I think it is a bit easier to read and follow.

src/test/java/use_cases/create_new_month/NewMonthUCITest.java
```diff
@@ -0,0 +1,101 @@
...
+        NewMonthID inputData = new NewMonthID(sessionStorage,monthID2,budgetValue);
+
+        NewMonthOD output = interactor.createOutput(inputData);
+        assertTrue(output.isSuccessful());
+        assertEquals(monthID2, sessionStorage.getMonthlyData(monthID2).getMonthID());
+        assertEquals(recurring,sessionStorage.getMonthlyData(monthID2).getExpenseData().get(0));
+    }
+
+    /**
+     * Test case when new MonthlyStorage is not added to session because there were duplicate
```
katarinaavucic, line 67: For these tests, you should be using assert.throws() instead of just saying that the EntityException will be thrown. You should only be allowing it to be thrown in correct cases (where you know it won't be thrown).

katarinaavucic (changes requested):
Overall really good! I just have some comments on the Javadoc as well as the protection of some methods and attributes.

src/main/java/views/monthly_menu/MonthMenuV.java
```diff
@@ -201,4 +200,13 @@ public void actionPerformed(ActionEvent event) {
             // call generateSummary
         }
     }
+
+    /**
+     * Open Month Menu and notify user.
+     * @param message notify user that Month Menu is updated
+     */
+    public void openMonthMenu(String message) {
```
aricasas, line 207: Do you think you could make the message optional, kinda like how I did it in openMainMenu? Because when I call this from the main menu I think having a popup every time you select a month would be kind of annoying to the user

aricasas (commented):
Apart from the one comment I left, I didn't see any other things to change that Kat didn't catch already. Nice job!

src/main/java/use_cases/create_new_month/NewMonthOD.java
YinLingithub: I didn't notice at all!! Will fix

src/main/java/views/create_new_month/NewMonthC.java
YinLingithub: I've been using controller (and for other, input data, use case interactor, etc.) in full to make it more descriptive.

src/main/java/views/create_new_month/NewMonthC.java
YinLingithub: I believe its implicitly public by default.

src/main/java/views/create_new_month/NewMonthV.java
YinLingithub: If you are referring to actionPerformed() and loadMonthMenu() they are public because they are implemented from interfaces. 

About the interface containing loadMonthMenu(), I thought Ari would also have this method to open the Month Menu, so I made a interface for it. However, I'm not very sure. @aricasas What do you think?

src/main/java/views/create_new_month/NewMonthV.java
YinLingithub: I'll hold on and change this later, as I might change loadMonthMenu() to private later. But thanks for telling me that, I didn't know!

src/test/java/use_cases/create_new_month/NewMonthUCITest.java
YinLingithub: The first two test are success cases so I know they won't be thrown. I'm editing the third test case so it won't do that!

src/main/java/views/monthly_menu/MonthMenuV.java
YinLingithub: You are right, it would be kind of annoying. I'll remove the message completely. I don't think I could make it optional.

YinLingithub: Also, I just realized the line 
_System.out.println(session);_ 
I don't think I meant to write this. I'll remove it.

src/main/java/views/create_new_month/NewMonthV.java
aricasas: I think they were referring to the instance attributes

aricasas: But regarding loadMonthMenu, I will use it to open the menus, but it doesn't really need to be in an interface. It technically could just be a public method and it still would work fine, but I would say keep it in an interface so that we're very explicit about keeping the method signature a certain way because other classes are calling it

src/main/java/views/monthly_menu/MonthMenuV.java
aricasas: The way I did it was that if you call the function with null for the message it will check for that and not do the popup

src/main/java/views/create_new_month/NewMonthV.java
YinLingithub: Ah, that make sense. I thought Kat was referring to the methods.

src/main/java/views/monthly_menu/MonthMenuV.java
aricasas: But if you want to remove it completely then that's fine too. Whichever you prefer

src/main/java/views/create_new_month/NewMonthC.java
YinLingithub: We could do a reformatting when we have everything finished so that everything looks consistent.

src/main/java/views/monthly_menu/MonthMenuV.java
YinLingithub: Actually I'll put it back. I made it the same way you did it.

src/main/java/views/create_new_month/NewMonthV.java
YinLingithub: After discussing about the interface with Ari, I decided that I will keep the interface, so I've moved loadMonthMenu() before the private method!

src/main/java/views/create_new_month/NewMonthC.java
katarinaavucic: Okay, sounds good!

src/main/java/views/create_new_month/NewMonthC.java
katarinaavucic: Alright! I think most of them are set public but yeah we could go back over and check. We do get marked for consistency though, so ideally we should do that by submission tomorrow.

src/main/java/views/create_new_month/NewMonthV.java
katarinaavucic: Sounds good!

katarinaavucic (approved):
Everything looks good! Just remember to go back over the default stuff for consistency later!

aricasas (approved):
Looks good

luluwuwu (approved):
approved.

PR #42, opened 2023-08-06
Fixed Add/Edit Expenses feature 

### LINKED ISSUE #41: Fix bugs in add/edit expense (recurring)

luluwuwu:
fixed bugs for recurring expenses, category assignments for expenses etc.

### DIFF
--- a/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
+++ b/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
@@ -69,8 +69,8 @@ public class ExpenseUCI implements ExpenseIB {
             }else{
                 Expense newExpense = new Expense(expenseID.getName(), selectedCategory,valueDouble);
                 month.addExpense(newExpense);}
-            ExpenseOD expenseODSuccessAdd = new ExpenseOD("You have added a new expense!");
-            return expenseOB.success(expenseODSuccessAdd);
+                ExpenseOD expenseODSuccessAdd = new ExpenseOD("You have added a new expense!");
+                return expenseOB.success(expenseODSuccessAdd);
 
         } catch (NumberFormatException e) {
             // NumberFormatException|NullPointerException fail: User tries to edit Expense value to an invalid number.
@@ -82,7 +82,7 @@ public class ExpenseUCI implements ExpenseIB {
             return expenseOB.fail(expenseODFailAdd);
         } catch(NoSuchElementException e){
         // NoSuchElementException fail: User tries to assign a category that does not exist in current month.
-        ExpenseOD expenseODFailEdit = new ExpenseOD("There is no such category in the current month. Please add a new expense or select existing expense!");
+        ExpenseOD expenseODFailEdit = new ExpenseOD("There is no such category in the current month. Please add a new category or select existing category!");
         return expenseOB.fail(expenseODFailEdit);}}
 
     /**
@@ -162,7 +162,7 @@ public class ExpenseUCI implements ExpenseIB {
      * @param monthCategoryData An ArrayList of categories.
      * @param name Category name.
      * @return Category with given String name.
-     * @throws NoSuchElementException thrown when couldn't find Expense with String name.
+     * @throws NoSuchElementException thrown when couldn't find Category with String name.
      */
     public Category findCategory (ArrayList < Category > monthCategoryData, String name) throws NoSuchElementException {
             for (Category category : monthCategoryData) {
--- a/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
+++ b/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
@@ -146,15 +146,17 @@ public class ExpenseUCI implements ExpenseIB {
 
     /**
      * Finds Expense with String name from a provided list of Expenses.
-     * @param ExpenseData An ArrayList of expenses.
+     * @param expenseData An ArrayList of expenses.
      * @param name Expense name.
      * @return Expense with given String name.
      * @throws NoSuchElementException thrown when couldn't find Expense with String name.
      */
-        public Expense findExpense(ArrayList<Expense> ExpenseData, String name)throws NoSuchElementException{
-            for (Expense expense : ExpenseData) {
-                if (Objects.equals(expense.getName(), name)) {
-                    return expense;}}
+        public Expense findExpense(ArrayList<Expense> expenseData, String name)throws NoSuchElementException{
+            if(expenseData != null) {
+                for (Expense expense : expenseData) {
+                    if (Objects.equals(expense.getName(), name)) {
+                        return expense;}}
+            }
             throw new NoSuchElementException();}
 
     /**
@@ -187,11 +189,11 @@ public class ExpenseUCI implements ExpenseIB {
      */
     private boolean changeInRecurringInfo() {
         if(expenseID.getIsRecurringExpense()) {
-            try {findExpense(recurringExpenseList, expenseID.getName());
+            try {findExpense(recurringExpenseList, expenseID.getOldExpense());
                 return false;
             }catch (NoSuchElementException e){return true;}
         }else{
-            try{findExpense(recurringExpenseList, expenseID.getName());
+            try{findExpense(recurringExpenseList, expenseID.getOldExpense());
                 return true;
             }catch(NoSuchElementException e) {
                 return false;}}}
--- a/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
+++ b/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
@@ -10,7 +10,6 @@ public class ExpenseUCI implements ExpenseIB {
     private final ExpenseOB expenseOB;
     private ExpenseID expenseID;
     private ArrayList<Expense> recurringExpenseList;
-    private ArrayList<Expense> monthExpenseList;
 
 
     /**
@@ -36,8 +35,8 @@ public class ExpenseUCI implements ExpenseIB {
      * Overrides method in ExpenseIB.
      * Attempts to add an expense with information from ExpenseID and returns a ExpenseOD indicating whether fail/success after execution.
      * Provides detailed fail messages according to each condition listed below:
-     * 1. NumberFormatException: User tries to add a new budget value that can not be converted to a double.
-     * 2. Expense budget less than 0: User tries to add a new budget value that is a negative number.
+     * 1. NumberFormatException: User tries to add a new value value that can not be converted to a double.
+     * 2. Expense value less than 0: User tries to add a new value value that is a negative number.
      * 3. EntityException: User tries to add an invalid Expense name but failed. (See entities/EntityException.java)
      * 4. Same recurring expense name: User tries to add an expense name that is a saved recurring expense in the current SessionStorage.
      *
@@ -58,12 +57,28 @@ public class ExpenseUCI implements ExpenseIB {
             double valueDouble = toDouble(expenseIDAdd.getValue());
 
             Category selectedCategory = findCategory(categoryList, expenseID.getOldCategory());
-            Category other = findCategory(categoryList, "Other");
 
             if (valueDouble < 0) {
-                // 2. Expense budget less than 0 fail
+                // 2. Expense value less than 0 fail
                 ExpenseOD expenseODFailAdd = new ExpenseOD("Expense value can't be less than $0. Please try again!");
                 return expenseOB.fail(expenseODFailAdd);
+            }
+            if (expenseID.getIsRecurringExpense()) {
+                if(recurringExpenseList.size() > 0) {
+                    for (Expense expense1 : recurringExpenseList) {
+                        if (expense1.getName().equals(expenseID.getName())) {
+                            //Repeated recurring expense name fail: When a user tries to add new recurring expense but the name already exist as a recurring expense.
+                            ExpenseOD expenseODFailAdd = new ExpenseOD("There is a recurring expense with this name, you don't need to add expense in month. " +
+                                    "(If this is not the same expense, please use another name!)");
+                            return expenseOB.fail(expenseODFailAdd);
+                        }
+                    }
+                }else{
+                    Expense newrecurExpense = new Expense(expenseID.getName(), selectedCategory, valueDouble);
+                    session.addRecurExpense(newrecurExpense);
+                    month.addExpense(newrecurExpense);
+                    ExpenseOD expenseODSuccessAdd = new ExpenseOD("You have created a new recurring expense!");
+                    return expenseOB.success(expenseODSuccessAdd);}
             }
             for (Expense expense1 : monthExpenseList) {
                 if (expense1.getName().equals(expenseID.getName())) {
@@ -73,23 +88,6 @@ public class ExpenseUCI implements ExpenseIB {
                 }
             }
 
-            if (expenseID.getIsRecurringExpense()) {
-                if(recurringExpenseList.size() > 0) {
-                    for (Expense expense1 : recurringExpenseList) {
-                        if (expense1.getName().equals(expenseID.getName())) {
-                            //Repeated name in recurringData
-                            ExpenseOD expenseODFailAdd = new ExpenseOD("There is a recurring expense with this name, you don't need to add expense in month. " +
-                                    "(If this is not the same expense, please use another name!)");
-                            return expenseOB.fail(expenseODFailAdd);
-                        }
-                    }
-                }else{
-                    Expense newrecurExpense = new Expense(expenseID.getName(), other, valueDouble);
-                    session.addRecurExpense(newrecurExpense);
-                    ExpenseOD expenseODSuccessAdd = new ExpenseOD("You have created a new recurring expense!");
-                    return expenseOB.success(expenseODSuccessAdd);}
-                    }
-
             Expense newExpense = new Expense(expenseID.getName(), selectedCategory,valueDouble);
 
             month.addExpense(newExpense);
@@ -97,17 +95,17 @@ public class ExpenseUCI implements ExpenseIB {
             return expenseOB.success(expenseODSuccessAdd);
 
         } catch (NumberFormatException | NullPointerException e) {
-            //1. NumberFormatException/NullPointerException fail
+            // NumberFormatException|NullPointerException fail: User tries to edit Expense value to an invalid number.
             ExpenseOD expenseODFailAdd = new ExpenseOD("Expense value is needs to be a number. Please try again!");
             return expenseOB.fail(expenseODFailAdd);
 
         } catch (EntityException e) {
-            //3. EntityException fail
+            // EntityException fail: User tries to add an invalid Expense name but failed. (See entities/EntityException.java)
             ExpenseOD expenseODFailAdd = new ExpenseOD("There is already a expense with this new name in this month.");
             return expenseOB.fail(expenseODFailAdd);
         } catch(NoSuchElementException e){
-        //2. NoSuchElementException fail
-        ExpenseOD expenseODFailEdit = new ExpenseOD("There is no such expense in the current month. Please add a new expense or select existing expense!");
+        //2. NoSuchElementException fail: User tries to assign a category that does not exist in current month.
+        ExpenseOD expenseODFailEdit = new ExpenseOD("There is no such category in the current month. Please add a new expense or select existing expense!");
         return expenseOB.fail(expenseODFailEdit);
     }}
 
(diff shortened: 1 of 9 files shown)


src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -58,37 +57,36 @@ public ExpenseOD addExpenseInMonth(ExpenseID expenseIDAdd) throws EntityExceptio
...
             Category selectedCategory = findCategory(categoryList, expenseID.getOldCategory());
-            Category other = findCategory(categoryList, "Other");
 
             if (valueDouble < 0) {
-                // 2. Expense budget less than 0 fail
+                // 2. Expense value less than 0 fail
                 ExpenseOD expenseODFailAdd = new ExpenseOD("Expense value can't be less than $0. Please try again!");
                 return expenseOB.fail(expenseODFailAdd);
             }
-            for (Expense expense1 : monthExpenseList) {
```
katarinaavucic, line 68: This should not be deleted. This block of code checks if the expense name already exists in that month, and is applicable regardless of if they are choosing recurringExpense or not.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -58,37 +57,36 @@ public ExpenseOD addExpenseInMonth(ExpenseID expenseIDAdd) throws EntityExceptio
...
-                if (expense1.getName().equals(expenseID.getName())) {
-                    //Repeated name in month
-                    ExpenseOD expenseODFailEdit = new ExpenseOD("There is already a expense with this new name in this month.");
-                    return expenseOB.fail(expenseODFailEdit);
-                }
-            }
-
             if (expenseID.getIsRecurringExpense()) {
                 if(recurringExpenseList.size() > 0) {
```
katarinaavucic: You don't need this entire if statement. Recurring expenses are recurring because they occur in the monthly storage of EVERY month. So we don't need this if statement, since if the new name is unique in that month, it is unique in the recurring expense list too.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -58,37 +57,36 @@ public ExpenseOD addExpenseInMonth(ExpenseID expenseIDAdd) throws EntityExceptio
...
+            }
+            for (Expense expense1 : monthExpenseList) {
+                if (expense1.getName().equals(expenseID.getName())) {
+                    //Repeated name in month
+                    ExpenseOD expenseODFailEdit = new ExpenseOD("There is already a expense with this new name in this month.");
+                    return expenseOB.fail(expenseODFailEdit);
+                }
+            }
 
             Expense newExpense = new Expense(expenseID.getName(), selectedCategory,valueDouble);
```
katarinaavucic: Everything else above the catch statements should be in an else block for clarity.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -131,76 +129,75 @@ public ExpenseOD editExpenseInMonth(ExpenseID expenseIDEdit) throws EntityExcept
         MonthlyStorage month = session.getMonthlyData(expenseID.getMonthID());
         ArrayList<Expense> monthExpenseList = month.getExpenseData();
         ArrayList<Category> categoryList = month.getCategoryData();
-        Expense selectedExpense;
 
         try {
             Category selectedCategory = findCategory(categoryList, expenseID.getOldCategory());
```
katarinaavucic: I would rename this, considering "getOldCategory" actually gets the new category of the expense.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -131,76 +129,75 @@ public ExpenseOD editExpenseInMonth(ExpenseID expenseIDEdit) throws EntityExcept
...
+                if(checkHaveSameNameInList(monthExpenseList, expenseID.getName())) {
+                    // Repeated name fail: When a user tries to edit the expense name to a existing expense in month.
                     ExpenseOD expenseODFailEdit = new ExpenseOD("There is already a expense with this new name in this month.");
                     return expenseOB.fail(expenseODFailEdit);
                 }}
-            if (changeInRecurringInfo()) {// 4. Same recurring expense name fail
                 if (expenseID.getIsRecurringExpense()) {
-                    if (checkHaveSameNameInList(recurringExpenseList)) {
+                    //Repeated recurring expense name fail: When a user tries to edit the current expense to a recurring expense but the name already exist as a recurring expense.
+                    if (checkHaveSameNameInList(recurringExpenseList,expenseID.getName())) {
```
katarinaavucic: This is checked above when you are checking the expenses in the month!

katarinaavucic: So you can delete everything in this if statement, also we don't need that output data message since its redundant to the one above about the same name.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -131,76 +129,75 @@ public ExpenseOD editExpenseInMonth(ExpenseID expenseIDEdit) throws EntityExcept
...
+            }
             if(!Objects.equals(expenseIDEdit.getName(), expenseIDEdit.getOldExpense())){
-                if(checkHaveSameNameInList(monthExpenseList)) {
-                    //1. Repeated name fail
+                if(checkHaveSameNameInList(monthExpenseList, expenseID.getName())) {
+                    // Repeated name fail: When a user tries to edit the expense name to a existing expense in month.
                     ExpenseOD expenseODFailEdit = new ExpenseOD("There is already a expense with this new name in this month.");
                     return expenseOB.fail(expenseODFailEdit);
                 }}
-            if (changeInRecurringInfo()) {// 4. Same recurring expense name fail
```
katarinaavucic, line 155: I actually think you could keep this line, it helps tell you whether or not there is a change, making other code redundant (like the name check in the recurring list can be removed because we already know the expense was not recurring before).

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -131,76 +129,75 @@ public ExpenseOD editExpenseInMonth(ExpenseID expenseIDEdit) throws EntityExcept
...
+                        session.addRecurExpense(selectedExpense);
                         selectedExpense.setCategory(selectedCategory);
-                        ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes of this expense to the category selected in current month!");
-                        return expenseOB.success(expenseODSuccessEditRecurring);
-                    }
-                }else {
+                        // Success edit to new recurring expense.
+                        ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes of this new recurring expense!");
+                        return expenseOB.success(expenseODSuccessEditRecurring);}
+                }else if(checkHaveSameNameInList(recurringExpenseList, expenseID.getOldExpense())&& !expenseIDEdit.getIsRecurringExpense()) {
```
katarinaavucic: You can just have an else statement here since we know that isRecurring is false and it used to be true (because we are keeping the if change in recurring statement), all we have to do is modify the changes and delete the expense from recurring.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -131,76 +129,75 @@ public ExpenseOD editExpenseInMonth(ExpenseID expenseIDEdit) throws EntityExcept
...
-                    ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes of this expense to a recurring expense in current session!");
+                    session.deleteRecurExpense(expenseID.getName());
+                    selectedExpense.setCategory(selectedCategory);
+                    //Success edit to remove recurring expense.
+                    ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes of this expense and it is no longer a recurring expense!");
                     return expenseOB.success(expenseODSuccessEditRecurring);
                 }
-                    }
 
             selectedExpense.setName(expenseIDEdit.getName());
```
katarinaavucic: For clarity, everything above the catch should be in an else clause.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -227,33 +225,14 @@ public Category findCategory (ArrayList < Category > monthCategoryData, String n
             throw new NoSuchElementException();
         }
 
-    /**
-     * Identifying the only two cases when Session.recurringExpenseData needs to updated when attempting to edit Expense.
-     * (The user changes their response for the isRecurringExpense checkbox)
-     * @return boolean suggesting there is a need to update Session.recurringExpenseData
-     */
-    private boolean changeInRecurringInfo() {
```
katarinaavucic, line 235: Please keep this I think this method is really great and helps keep code simpler in the main method!!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: Also, notice how you have the same three lines of code repeated wherever you are editing an expense? You could make an editExpense() private helper method with those three set_ lines and that would make the code up here cleaner!

katarinaavucic (changes requested):
Overall it is looking better and you had a lot of the same ideas as I did! There are some issues with the way you are checking recurring expenses still, so I tried to clarify as best as I could. Also here is the general structure I was thinking for the code in pseudocode:

try clause in add:
if valuedouble...
for expense in monthExpenseList...
if is recurring... (we need to create expense and add to month and session)
else... (we need to create expense and add to month only)
** alternatively you could have the if recurring after you create and add to month so that if its recurring it will add to session.

try clause in edit:
if valuedouble...
if equals getName oldExpenseGetName...
if change in recurring...
     if is recurring.... (we know that expense should be edited and added to recurExpense)
     else... (we know that expense should be edited and removed from recurExpense)
else (we know expense only needs to be edited)

Hope that helps!

src/main/java/views/add_edit_epense_views/AddExpenseV.java
YinLingithub, line 1: Line 71. Set HIDE_ON_CLOSE

src/main/java/views/add_edit_epense_views/AddExpenseV.java
```diff
@@ -24,14 +23,15 @@ public class AddExpenseV extends Component implements ActionListener, LoadMonthM
     private final JComboBox<String> categoryCombo;
     private final String selectedExpense;
     private String selectedCategory;
-    private JCheckBox isRecurringCheckBox;
+    private final JCheckBox isRecurringCheckBox = new JCheckBox("This is a recurring expense.");
```
YinLingithub: I was testing the view, and I think having a whole sentence commenting the check box may be too long. Make it shorter, like "Is recurring expense".

src/main/java/views/add_edit_epense_views/EditExpenseV.java
```diff
@@ -33,18 +33,20 @@ public class EditExpenseV extends Component implements ActionListener, LoadMonth
```
YinLingithub, line 34: Same here, maybe make the checkbox comment shorter.

src/main/java/views/add_edit_epense_views/EditExpenseV.java
YinLingithub, line 1: Line 73 or 75: HIDE_ON_CLOSE

YinLingithub (commented):
You should also add
`panel.add(submit);`
in openEditCategory() of EditCategoryV (around line 70)

src/main/java/views/add_edit_epense_views/AddExpenseV.java
```diff
@@ -24,14 +23,15 @@ public class AddExpenseV extends Component implements ActionListener, LoadMonthM
     private final JComboBox<String> categoryCombo;
```
YinLingithub, line 24: You would need to extend JFrame instead of Component for your JOptionPane.showMessageDialog() to work. This also applies to EditExpenseV, AddCategoryV, EditCategoryV. 

Since we have all classes merged now, you should run Main.java to test your view.

YinLingithub:
This is what you need to change in AddCategoryV and EditCategoryV. 

[https://github.com/CSC207-2023Y-UofT/course-project-DoorKeepers/commit/6dd16641f436d9759cb74b06ac76236ad9ff4e5b](url)

[https://github.com/CSC207-2023Y-UofT/course-project-DoorKeepers/commit/4055e63155e81a7d5680e985675a87d07acc2dbe](url)

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: Got it. It took me a while to think about what to write for this failed message and it still turns out awkward, so actually kinda glad that we won't use it. ahah

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: What you're saying makes sense.

luluwuwu: They're just moved under from lines 82-89

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: I think I'll just have the ExpenseOD instantiation and the return outside, it shows an error for not having a return statement, and we know it's not part of the exceptions that we're dealing with!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: Yup, I saw! but instead the should be checked where I specified above, since if they are checked below you are not checking the recurring expenses (you are checking in the recurExpense list which is redundant, so if you move this code block back it works better!).

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: I'm refactoring all of them to "category". (getCategory() etc. )

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: LOL no worries! Honestly, this task was pretty daunting so I understand youuu.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: Sounds good!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: Oh yes

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: That sounds like a good idea

src/main/java/views/add_edit_epense_views/AddExpenseV.java
luluwuwu: right tks

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: Nice!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: Nvm the last if statement has an else, so these lines will have to be independent

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: Thanks! :D you could probably do a similar thing for adding as well!

luluwuwu: I looked at it for addExpense, but it will just be two lines of code and it will have to pass in the same amount of parameters. So I think I'll just leave it as is! (Also one of them will be MonthlyStorage.addeEpense() which is going to throw an entityException and I will like to avoid it at all costs lol)

katarinaavucic: Sounds good!! Omg real the EntityException there has gotten so annoying so yeah best to leave it out!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: I see what the confusion is here, I am suggesting to move the for loop back where it was (it needs to be there otherwise we arent checking for the same name in the recurring expenses) and then have this code go in the else block that comes after the if is recurring expense. Like how i said it in the overall comment! But i think the reason why you are saying there is no else is because the for loop is between the code (unless I am misunderstanding!).

luluwuwu: No yah you're right I moved the previous lines into the wrong else, now they are in the else.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -53,70 +47,47 @@ public ExpenseOD addExpenseInMonth(ExpenseID expenseIDAdd) throws EntityExceptio
...
-                // 2. Expense value less than 0 fail
+                //Expense value less than 0 fail: When a user tries to add the expense value with a negative number.
                 ExpenseOD expenseODFailAdd = new ExpenseOD("Expense value can't be less than $0. Please try again!");
-                return expenseOB.fail(expenseODFailAdd);
-            }
+                return expenseOB.fail(expenseODFailAdd);}
             if (expenseID.getIsRecurringExpense()) {
                 if(recurringExpenseList.size() > 0) {
-                    for (Expense expense1 : recurringExpenseList) {
+                    for (Expense expense1 : monthExpenseList) {
```
katarinaavucic: This for loop should be above line 56. Currently it is only checking for if an expense has the same name if the expense is recurring and there are recurring expenses, which should not happen. Additionally, you should get rid of the if statement in line 57, because it prevents recurring expenses from being added if no recurring expenses exist, which means the user would never be able to add a recurring expense.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -130,53 +101,34 @@ public ExpenseOD editExpenseInMonth(ExpenseID expenseIDEdit) throws EntityExcept
...
-                }else if(checkHaveSameNameInList(recurringExpenseList, expenseID.getOldExpense())&& !expenseIDEdit.getIsRecurringExpense()) {
-                    selectedExpense.setName(expenseIDEdit.getName());
-                    expenseIDEdit.setValue(valueDouble);
-                    selectedExpense.setValue(valueDouble);
+                    setEditInfo(expenseID, valueDouble, selectedExpense,selectedCategory);
+                    session.addRecurExpense(selectedExpense);
+                    // Success edit to new recurring expense.
+                    ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes of this new recurring expense!");
+                    return expenseOB.success(expenseODSuccessEditRecurring);
+                }else if (checkHaveSameNameInList(recurringExpenseList, expenseID.getOldExpense())){
```
katarinaavucic: This shouldn't be an else if, it should just be else! Otherwise this block won't run because it has already been failed above when checking for the same name. But other than that minor change this method is good!

katarinaavucic: This else runs when the expense was recurring but now is not! That's why we are deleting it from recurExpense.

katarinaavucic (changes requested):
YAYYYY I think we are almost at the non-buggy implementation!!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: You're right

katarinaavucic (dismissed):
YAYYYYYYY!!! It should work now, just verify that it runs on all of your tests :D.

katarinaavucic (approved):
YAYY everything works!

YinLingithub (approved):
Good job, Lulu and Kat!!!

aricasas (approved):
Nice job!

PR #56, opened 2023-08-12
Implementation of Factory Design Pattern in Entities

### LINKED ISSUE #55: Implementation of Factory Design Pattern in Entities
Implementing the part of the factory design pattern that resides in the entities.

katarinaavucic:
I implemented the Factory Design Pattern in the Entities so that the ExpenseUseCaseInteractor and CategoryUseCaseInteractor can create one of the four concrete MonthObjectFactoryInputData classes (according to the objective) and call either the createMonthObject() or editMonthObject() method of the correct concrete MonthObjectFactory class with this object.

This adheres to the Single Responsibility Principle in SOLID, as the UseCaseInteractor can focus on error handling and creating the OutputData and leave the creation to another class. It also addresses the problem where the UseCaseInteractor is constructing the entities directly, which can cause difficulties when extending the program.

I also included some tests to demonstrate how the classes work.


src/main/java/entities/MonthObject.java
```diff
@@ -2,6 +2,6 @@
 
 /**
  * An interface that objects belonging to the MonthlyStorage implement. Expense and Category implement this so that
- * they can be created by MonthObjectCreator classes.
+ * they can be created by MonthObjectFactory classes.
  */
 public interface MonthObject {}
```
YinLingithub: The interface does not contain any methods? Shouldn't it have methods like getName() setName() getValue() setValue() so they could be called in the use case interactor?

src/main/java/entities/MonthObject.java
YinLingithub, line 1: The changes to Category are not shown here. I think you should fetch and update main merge main to you branch to keep things look the way they are.

YinLingithub: If any changes got lost in the process, you should copy them from the mistake commit in main.

src/main/java/entities/MonthObject.java
luluwuwu: Maybe it'll be some getters and setters for the mutual items such as name and value, that will be required for the classes that implement MonthObject (in our case Expense and Category). This in turn can be used in the use case interactor of the Add/Edit features!

luluwuwu (commented):
Overall very concise and understandable just remember to implement MonthObject interface as there is a way to not leave it empty and become a codesmell!

src/main/java/entities/MonthObject.java
katarinaavucic: Hmm that was something I thought about, but since value is budget in Category, I decided against it. I think i will add name to both, since we can agree those are the same in both Category and Expense, and every MonthObject should have a name!

src/test/java/entities/ExpenseFactoryTest.java
```diff
@@ -0,0 +1,102 @@
...
+        ExpenseEditorInputData inputData = new ExpenseEditorInputData("Subway", c1, 10, e1);
+        MonthObjectFactory editor = new ExpenseFactory();
+        MonthObject monthObject = editor.editMonthObject(inputData);
+
+        Expense expense = (Expense) monthObject;
+        Assertions.assertEquals(expense.getName(), inputData.getName());
+        Assertions.assertEquals(expense.getCategory(), inputData.getCategory());
+        Assertions.assertEquals(expense.getValue(), inputData.getValue());
+        Assertions.assertEquals(expense, inputData.getExpense());
+    }
```
YinLingithub: Is this test needed? Is so, wouldn't we also need a similar test for Category editing?

luluwuwu: I would say we can get rid of this one since there is no related fail messages returned in add/edit features. Such that when there are no modifications, the UCI handles it as a success edit!

aricasas (dismissed):
The factory implementations seem good! Only thing I noticed is the same test case that Yin commented on. I'm not sure if we really need to test for editing to the same values as a specific test case, since that should be covered by the general editing tests right?

aricasas:
Wait I just realized but shouldn't the factory input data classes have setters for all the fields? So that you don't have to pass in everything through a single constructor call. Because otherwise what would be the point of having a factory if the way to build the factory input data is the exact way we used to build the objects themselves?

So that you can do something like this:
```
ExpenseCreatorInputData expenseCreatorInputData = new ExpenseCreatorInputData();
expenseCreatorInputData.setName(name1);
expenseCreatorInputData.setCategory(category1);
expenseCreatorInputData.setValue(100);

Expense newExpense = factory.createMonthObject(expenseCreatorInputData);
``` 

So that in Lulu's code instead of setting a bunch of instance attributes in her UCI class, she would set the values in the factory input data directly

src/test/java/entities/ExpenseFactoryTest.java
katarinaavucic: I added it because of what Lulu is saying here, since we allow the user to edit nothing it is considered a success case and thus would be a possible input for the ExpenseFactory. It is not a useful test per say but I did want to include one test case for every possible option the user could do. If you feel it is unnecessary I can remove it and if you feel it is okay to stay I will add another one in the CategoryFactoryTest!

katarinaavucic:
> Wait I just realized but shouldn't the factory input data classes have setters for all the fields? So that you don't have to pass in everything through a single constructor call. Because otherwise what would be the point of having a factory if the way to build the factory input data is the exact way we used to build the objects themselves?
> 
> So that you can do something like this:
> 
> ```
> ExpenseCreatorInputData expenseCreatorInputData = new ExpenseCreatorInputData();
> expenseCreatorInputData.setName(name1);
> expenseCreatorInputData.setCategory(category1);
> expenseCreatorInputData.setValue(100);
> 
> Expense newExpense = factory.createMonthObject(expenseCreatorInputData);
> ```
> 
> So that in Lulu's code instead of setting a bunch of instance attributes in her UCI class, she would set the values in the factory input data directly

I see what you are saying here. The MonthObjectFactoryInputData class is not a part of the Factory design pattern, it is something I made to accommodate for the fact that the addMonthObject and editMonthObject methods need different parameters in the concrete ExpenseFactory and CategoryFactory classes. This way the abstract methods in MonthObjectFactory could take one abstract parameter MonthObjectFactoryInputData. Thus, any child class of this abstract input data class could be passed in, so we could accommodate for the differences in factory parameters.

If Lulu feels your suggestion would improve her code, I could definitely fix that, although since it is not technically part of the Factory design pattern I do not think it invalidates the functionality of it.

src/test/java/entities/ExpenseFactoryTest.java
aricasas: Oh ok that makes sense. It is technically a bit redundant, but it is okay since it is explicitly handling a possible user input. I'd say to add one to category too then. But maybe you could explain in the JavaDoc why you decided to test this specific case explicitly?

aricasas:
> > Wait I just realized but shouldn't the factory input data classes have setters for all the fields? So that you don't have to pass in everything through a single constructor call. Because otherwise what would be the point of having a factory if the way to build the factory input data is the exact way we used to build the objects themselves?
> > So that you can do something like this:
> > ```
> > ExpenseCreatorInputData expenseCreatorInputData = new ExpenseCreatorInputData();
> > expenseCreatorInputData.setName(name1);
> > expenseCreatorInputData.setCategory(category1);
> > expenseCreatorInputData.setValue(100);
> > 
> > Expense newExpense = factory.createMonthObject(expenseCreatorInputData);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So that in Lulu's code instead of setting a bunch of instance attributes in her UCI class, she would set the values in the factory input data directly
> 
> I see what you are saying here. The MonthObjectFactoryInputData class is not a part of the Factory design pattern, it is something I made to accommodate for the fact that the addMonthObject and editMonthObject methods need different parameters in the concrete ExpenseFactory and CategoryFactory classes. This way the abstract methods in MonthObjectFactory could take one abstract parameter MonthObjectFactoryInputData. Thus, any child class of this abstract input data class could be passed in, so we could accommodate for the differences in factory parameters.
> 
> If Lulu feels your suggestion would improve her code, I could definitely fix that, although since it is not technically part of the Factory design pattern I do not think it invalidates the functionality of it.

Ok. If it doesn't invalidate the design pattern, it should be fine then. I think might not be getting the point of having the factory then. Ithink I have to re-read the notes and refactoring guru lol.

src/test/java/entities/ExpenseFactoryTest.java
YinLingithub: Makes sense. And this reminded me of another potential problem: 

> If the user successfully edits a Category from name A to B, and then edits a expense that has Category type A while not changing name, value, and Category. Then it will not b successful because the Category now has name B. 

@katarinaavucic @luluwuwu 
I'm wondering if we have anything that prevents this, if not, it should probably by implemented here too. (One solution I could think of is to create a new Category with name B, and edit all expenses with Category A to having Category B. Then delete Category with name A.)

luluwuwu: This should not cause a problem: A Category Object rather than just the category name is passed in as the ExpenseID. So when a user edits a category name that had expenses belonging to this category,  any expense.getCategory() should return the Category object with the new name!

YinLingithub: Right! Thanks! I got totally confused... 
Then just as Ari suggested, a similar test case should be added for Category, and I believe there's no more problems!

katarinaavucic: Perfect, I will implement one in Category and add info to the Javadoc then!

katarinaavucic:
> > > Wait I just realized but shouldn't the factory input data classes have setters for all the fields? So that you don't have to pass in everything through a single constructor call. Because otherwise what would be the point of having a factory if the way to build the factory input data is the exact way we used to build the objects themselves?
> > > So that you can do something like this:
> > > ```
> > > ExpenseCreatorInputData expenseCreatorInputData = new ExpenseCreatorInputData();
> > > expenseCreatorInputData.setName(name1);
> > > expenseCreatorInputData.setCategory(category1);
> > > expenseCreatorInputData.setValue(100);
> > > 
> > > Expense newExpense = factory.createMonthObject(expenseCreatorInputData);
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > So that in Lulu's code instead of setting a bunch of instance attributes in her UCI class, she would set the values in the factory input data directly
> > 
> > 
> > I see what you are saying here. The MonthObjectFactoryInputData class is not a part of the Factory design pattern, it is something I made to accommodate for the fact that the addMonthObject and editMonthObject methods need different parameters in the concrete ExpenseFactory and CategoryFactory classes. This way the abstract methods in MonthObjectFactory could take one abstract parameter MonthObjectFactoryInputData. Thus, any child class of this abstract input data class could be passed in, so we could accommodate for the differences in factory parameters.
> > If Lulu feels your suggestion would improve her code, I could definitely fix that, although since it is not technically part of the Factory design pattern I do not think it invalidates the functionality of it.
> 
> Ok. If it doesn't invalidate the design pattern, it should be fine then. I think might not be getting the point of having the factory then. Ithink I have to re-read the notes and refactoring guru lol.

I understand why you were confused! I will add the reasoning for the input data classes in their Javadocs to prevent further confusion.

luluwuwu:
> > > > Wait I just realized but shouldn't the factory input data classes have setters for all the fields? So that you don't have to pass in everything through a single constructor call. Because otherwise what would be the point of having a factory if the way to build the factory input data is the exact way we used to build the objects themselves?
> > > > So that you can do something like this:
> > > > ```
> > > > ExpenseCreatorInputData expenseCreatorInputData = new ExpenseCreatorInputData();
> > > > expenseCreatorInputData.setName(name1);
> > > > expenseCreatorInputData.setCategory(category1);
> > > > expenseCreatorInputData.setValue(100);
> > > > 
> > > > Expense newExpense = factory.createMonthObject(expenseCreatorInputData);
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > So that in Lulu's code instead of setting a bunch of instance attributes in her UCI class, she would set the values in the factory input data directly
> > > 
> > > 
> > > I see what you are saying here. The MonthObjectFactoryInputData class is not a part of the Factory design pattern, it is something I made to accommodate for the fact that the addMonthObject and editMonthObject methods need different parameters in the concrete ExpenseFactory and CategoryFactory classes. This way the abstract methods in MonthObjectFactory could take one abstract parameter MonthObjectFactoryInputData. Thus, any child class of this abstract input data class could be passed in, so we could accommodate for the differences in factory parameters.
> > > If Lulu feels your suggestion would improve her code, I could definitely fix that, although since it is not technically part of the Factory design pattern I do not think it invalidates the functionality of it.
> > 
> > 
> > Ok. If it doesn't invalidate the design pattern, it should be fine then. I think might not be getting the point of having the factory then. Ithink I have to re-read the notes and refactoring guru lol.
> 
> I understand why you were confused! I will add the reasoning for the input data classes in their Javadocs to prevent further confusion.

I would agree with Ari. If you go to my pull request on factory methods, the two setCreator/EditorInputInfo methods would be difficult to process/read due to lengthy instances to construct the MonthObjectFactoryInputData (since we're minimizing making private attributes.)

katarinaavucic:
> > > > > Wait I just realized but shouldn't the factory input data classes have setters for all the fields? So that you don't have to pass in everything through a single constructor call. Because otherwise what would be the point of having a factory if the way to build the factory input data is the exact way we used to build the objects themselves?
> > > > > So that you can do something like this:
> > > > > ```
> > > > > ExpenseCreatorInputData expenseCreatorInputData = new ExpenseCreatorInputData();
> > > > > expenseCreatorInputData.setName(name1);
> > > > > expenseCreatorInputData.setCategory(category1);
> > > > > expenseCreatorInputData.setValue(100);
> > > > > 
> > > > > Expense newExpense = factory.createMonthObject(expenseCreatorInputData);
> > > > > ```
> > > > > 
> > > > > 
> > > > >     
> > > > >       
> > > > >     
> > > > > 
> > > > >       
> > > > >     
> > > > > 
> > > > >     
> > > > >   
> > > > > So that in Lulu's code instead of setting a bunch of instance attributes in her UCI class, she would set the values in the factory input data directly
> > > > 
> > > > 
> > > > I see what you are saying here. The MonthObjectFactoryInputData class is not a part of the Factory design pattern, it is something I made to accommodate for the fact that the addMonthObject and editMonthObject methods need different parameters in the concrete ExpenseFactory and CategoryFactory classes. This way the abstract methods in MonthObjectFactory could take one abstract parameter MonthObjectFactoryInputData. Thus, any child class of this abstract input data class could be passed in, so we could accommodate for the differences in factory parameters.
> > > > If Lulu feels your suggestion would improve her code, I could definitely fix that, although since it is not technically part of the Factory design pattern I do not think it invalidates the functionality of it.
> > > 
> > > 
> > > Ok. If it doesn't invalidate the design pattern, it should be fine then. I think might not be getting the point of having the factory then. Ithink I have to re-read the notes and refactoring guru lol.
> > 
> > 
> > I understand why you were confused! I will add the reasoning for the input data classes in their Javadocs to prevent further confusion.
> 
> I would agree with Ari. If you go to my pull request on factory methods, the two setCreator/EditorInputInfo methods would be difficult to process/read due to lengthy instances to construct the MonthObjectFactoryInputData (since we're minimizing making private attributes.)

Sounds good! I will push the previous changes I made and modify these classes in the next commit!

YinLingithub (approved):
Looks good!

luluwuwu (approved):
PErfect!

aricasas (approved):
Everything looks great!

PR #58, opened 2023-08-12
57 fix edit feature

### LINKED ISSUE #57: FIx EditCategory Feature
Improved implementation to prevent unexpected fail messages for edit category feature.

luluwuwu:
The reason for the unexpected failed messages from failed edits is due to the implementation in the Controllers. 

For example: When a user wishes to edit a category but did not select a category in EditCategoryV, the selected category is null and will rather be classified as an addCategory use case in CategoryC. And a NoSuchElementException would be thrown in CategoryUCI addCategoryInMonth method and is mistaken for a NoSuchElementException thrown in the editCategoryInMonth method. 

Solution: Now this incident is handled in EditCategoryV and is not the responsibility of CategoryUCI. 
(vice versa when a user wishes to edit an expense)

### DIFF
--- a/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
+++ b/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
@@ -79,15 +79,7 @@ public class ExpenseUCI implements ExpenseIB {
         } catch (EntityException e) {
             // EntityException fail: User tries to add an invalid Expense name but failed. (See entities/EntityException.java)
             ExpenseOD expenseODFailAdd = new ExpenseOD("There is already a expense with this new name in this month.");
-            return expenseOB.fail(expenseODFailAdd);
-        } catch(NoSuchElementException e){
-            // NoSuchElementException fail: User tries to edit an expense but did not select an expense in editExpenseView.
-            //This exception for an EditExpenseUseCase is caught here due to the implementation in ExpenseC.
-            //The determining factor for entering an Add or Edit UseCase is by checking if the selected Expense is null,
-            // hence when a user did not select an expense when editing an expense, ExpenseC actually classifies the instance
-            // as an AddExpenseUseCase although the user is interacting with the editExpenseView.
-        ExpenseOD expenseODFailEdit = new ExpenseOD("Please select an expense you wish to edit.");
-        return expenseOB.fail(expenseODFailEdit);}}
+            return expenseOB.fail(expenseODFailAdd);}}
 
     /**
      * Attempts to edit an expense with information from ExpenseID and returns a ExpenseOD indicating whether fail/success after execution.
--- a/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
+++ b/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
@@ -81,8 +81,12 @@ public class ExpenseUCI implements ExpenseIB {
             ExpenseOD expenseODFailAdd = new ExpenseOD("There is already a expense with this new name in this month.");
             return expenseOB.fail(expenseODFailAdd);
         } catch(NoSuchElementException e){
-        // NoSuchElementException fail: User tries to assign a category that does not exist in current month.
-        ExpenseOD expenseODFailEdit = new ExpenseOD("There is no such category in the current month. Please add a new category or select existing category!");
+            // NoSuchElementException fail: User tries to edit an expense but did not select an expense in editExpenseView.
+            //This exception for an EditExpenseUseCase is caught here due to the implementation in ExpenseC.
+            //The determining factor for entering an Add or Edit UseCase is by checking if the selected Expense is null,
+            // hence when a user did not select an expense when editing an expense, ExpenseC actually classifies the instance
+            // as an AddExpenseUseCase although the user is interacting with the editExpenseView.
+        ExpenseOD expenseODFailEdit = new ExpenseOD("Please select an expense you wish to edit.");
         return expenseOB.fail(expenseODFailEdit);}}
 
     /**
@@ -136,7 +140,7 @@ public class ExpenseUCI implements ExpenseIB {
 
             } catch(NoSuchElementException e){
                 //NoSuchElementException fail: User tries to edit an expense that does not exist.
-                ExpenseOD expenseODFailEdit = new ExpenseOD("There is no such expense in the current month. Please add a new expense or select existing expense!");
+                ExpenseOD expenseODFailEdit = new ExpenseOD("Please make sure you have selected the expense you wish to edit or assigned a category to the expense!");
                 return expenseOB.fail(expenseODFailEdit);
             } catch(NumberFormatException e){
                 // NumberFormatException|NullPointerException fail: User tries to edit Expense value to an invalid number.
--- a/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
+++ b/src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
@@ -9,8 +9,6 @@ import java.util.Objects;
 public class ExpenseUCI implements ExpenseIB {
     private final ExpenseOB expenseOB;
     private ExpenseID expenseID;
-    private ArrayList<Expense> recurringExpenseList;
-
     /**
      * Constructs ExpenseUCI.
      *
@@ -45,7 +43,6 @@ public class ExpenseUCI implements ExpenseIB {
         MonthlyStorage month = expenseIDAdd.getSession().getMonthlyData(expenseIDAdd.getMonthID());
         ArrayList<Expense> monthExpenseList = month.getExpenseData();
         ArrayList<Category> categoryList = month.getCategoryData();
-        this.recurringExpenseList = session.getRecurData();
 
         try {double valueDouble = toDouble(expenseIDAdd.getValue());
             Category selectedCategory = findCategory(categoryList, expenseID.getCategory());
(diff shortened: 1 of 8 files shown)


katarinaavucic (approved):
Nice fix! Everything looks good to me.

katarinaavucic (dismissed):
Re-approving :).

src/main/java/views/add_edit_category_views/EditCategoryV.java
```diff
@@ -97,16 +98,11 @@ public void actionPerformed(ActionEvent evt) {
             else if (budgetInput.getText().isEmpty()){
                 JOptionPane.showMessageDialog(this,"Please enter the previous category budget if you don't wish to edit. Thanks.");
             }
-            // Check if user selects an old category.
-            else if (categoryCombo.getSelectedItem() == null) {
-                JOptionPane.showMessageDialog( this, "Please select a category to edit.");
-            }
             else {
                 tryUseCaseEdit();
             }
```
aricasas: This is just kind of a nitpick since it doesn't really affect how the code works, but why are these ifs embedded inside the else and not just continuing the if-else chain?

aricasas (dismissed):
Approved! Everything seems good to me. I just had one tiny comment in EditCategoryV that can be safely ignored if you like

src/main/java/views/add_edit_category_views/EditCategoryV.java
luluwuwu: It's because there wasn't the else if before and I didn't change the rest of my code lol. If you think that affects code reading, I can fix it! lmk

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
YinLingithub, line 1: I think there may still be problem updating the recuring expense info in your editExpenseInMonth. 

When I try to edit recuring expense, the edit view will show me that the expense is recurring and I could uncheck the checkbox and submit that as a change. Then when I make a new month, it will tell me that an error has occurred when initializing the new month. This means "two Expenese in recurData to have the same name", since I used a different year and month when creating the new month. 

I then tried to edit the same expense again. The view still showed me that it is a recurring expense, and it recognize me unchecking the checkbox as a valid change.

YinLingithub: This is actually pretty weird, because if I created a new month (say X) before editing the expense, the expense (say exp1)will be added to new month as recuring expense. And if I then edit the expense to make it non recurring (and change name to exp2), then month X recognize that this expense is non recurring, but the name is still exp1.

src/main/java/views/add_edit_epense_views/EditExpenseV.java
```diff
@@ -109,7 +106,10 @@ public void actionPerformed(ActionEvent evt) {
         // Set isRecurringCheckBox checked if the expense being edited was a recurring expense
         this.selectedCategory = (String) categoryCombo.getSelectedItem();
         this.isRecurring = isRecurringCheckBox.isSelected();
-        }else if(evt.getSource() == submit){
+        }else if(expenseCombo.getSelectedIndex() == -1){
```
YinLingithub: I think you would need such a test inside the if ext.getSource() ==submit clause, since when I tried to edit a expense without selecting a expense I want to submit (which shows as if i've selected the first expense in the list), it would just give NoSuchElementException. 

An alternative that I can guarantee it works, is to make the first element of the list a empty 
string "". This way, you can catch the "" case in the if submit clause, and it is also visually more understandable to the user. (This is how I implemented in the MonthMenuView for selecting a month input)

src/main/java/views/add_edit_epense_views/EditExpenseV.java
```diff
@@ -122,7 +122,6 @@ public void actionPerformed(ActionEvent evt) {
                 JOptionPane.showMessageDialog( this, message.getMessage());}
```
YinLingithub: Line 117:

should be frame.setVisible(false)

src/main/java/views/add_edit_category_views/EditCategoryV.java
```diff
@@ -87,8 +87,9 @@ public void actionPerformed(ActionEvent evt) {
         //Two ActionListeners with different behaviours differentiated by checking evt.getSource().
         if (evt.getSource() == categoryCombo) {
             this.selectedCategory = (String) categoryCombo.getSelectedItem();
-        }
-        else {
+        } else if(categoryCombo.getSelectedIndex() == -1){
+            JOptionPane.showMessageDialog( this, "Please select a category you wish to edit.");
```
YinLingithub: Maybe you could try if it works without having this else if clause, because the same code didn't work in your EditExpenseV.

src/main/java/views/add_edit_epense_views/EditExpenseV.java
luluwuwu: I believe I fixed this error and I do agree it's a valid case for a test since it's prone to errors. However, I'm struggling to build a test, any suggestions?

src/main/java/views/add_edit_category_views/EditCategoryV.java
luluwuwu: I've abandoned this if statement to address the error, please take a look in my next commit.

YinLingithub: I see that it's been removed from EditExpenseV, but unfortunately it also doesn't work in here. When I tried to edit a category without clicking the dropdown list, it would just add a new month rather than warning the user to select a category. I think you could try replacing 
**line 23:**  
private String selectedCategory;
with
private String selectedCategory = "";

**line 90:** 
else if(categoryCombo.getSelectedIndex() == -1)
with
if(selectedCategory == "")

luluwuwu: That should be a problem in EditCategory. I realized I didn't have that one updated. I'm not sure what you meant by "it added you a new month", may you elaborate?

src/main/java/views/add_edit_epense_views/EditExpenseV.java
```diff
@@ -102,27 +104,36 @@ public void actionPerformed(ActionEvent evt) {
...
-            JOptionPane.showMessageDialog( this, "Please select an expense you wish to edit.");
-        }else if(evt.getSource() == submit){
-            String selectedCategory = (String) categoryCombo.getSelectedItem();
-            boolean isRecurring = isRecurringCheckBox.isSelected();
-            ExpenseOD message = null;
-            try {
-                message = controller.expenseInMonth(nameInput.getText(), valueInput.getText(), selectedCategory,
-                        isRecurring, monthID, currSession, selectedExpense);
-                // Update Month Menu
+        if (evt.getSource() == categoryCombo) {
```
luluwuwu: I know we talked about minimizing the use of actionListeners, but I don't know how to make it work without one in this case. Please provide suggestions if any!

src/main/java/views/add_edit_category_views/EditCategoryV.java
YinLingithub: **Yes, its a problem in EditCategory.** (Sorry it was a bit confusing. I said EditExpenseV you previously had similar code in EditExpenseV too.)

From my understanding, with this code you are trying to enforce the user to select a Category, and thereby separating this EditCategory case from AddCategory case. However, **there is a problem with this code, and it does not notify the user that they haven't selected a Category**. As a result, your controller handles this "have not selected a Category to edit" case as AddCategory case (because **the selectedCategory inputted is null**), and created a new Category instead. 

Hopefully the attached screenshots can give you an idea of what I'm talking about. 
Image 1: trys to edit without selecting any Category (the Other category is shown in the ComboBox part by default)
Screenshot 2023-08-13 171345
Image 2: result of submitting without selecting a Category to edit
Screenshot 2023-08-13 171400

src/main/java/views/add_edit_epense_views/EditExpenseV.java
```diff
@@ -102,27 +104,36 @@ public void actionPerformed(ActionEvent evt) {
                 if (recurExpense.getName().equals(selectedExpense)) {
                     // Set isRecurringCheckBox checked if the expense being edited was a recurring expense
                     isRecurringCheckBox.setSelected(true);}}
-        }
+        } else if (selectedExpense == null) {
+            JOptionPane.showMessageDialog(this, "Please select an expense you wish to edit.");}
```
YinLingithub: I think this should go inside the `if (evt.getSource() == categoryCombo)` clause.

src/main/java/views/add_edit_epense_views/EditExpenseV.java
```diff
@@ -102,27 +104,36 @@ public void actionPerformed(ActionEvent evt) {
...
-            ExpenseOD message = null;
-            try {
-                message = controller.expenseInMonth(nameInput.getText(), valueInput.getText(), selectedCategory,
-                        isRecurring, monthID, currSession, selectedExpense);
-                // Update Month Menu
+        if (evt.getSource() == categoryCombo) {
+            this.selectedCategory = (String) categoryCombo.getSelectedItem();
+        } else if (selectedCategory == null){
+            JOptionPane.showMessageDialog(this, "Please select the previous category if you " +
+                "don't wish to edit. Thanks.");}
```
YinLingithub: Same for this, it should go inside the `if (evt.getSource() == categoryCombo)` clause.

src/main/java/views/add_edit_category_views/EditCategoryV.java
YinLingithub: I think it's actually solved in your latest commit, so I'll mark this as resolved.

src/main/java/views/add_edit_epense_views/EditExpenseV.java
YinLingithub: You don't build a test for it because it's in the view. You actually run the main() to test it.

src/main/java/views/add_edit_epense_views/EditExpenseV.java
YinLingithub: resolve as not relevant

src/main/java/views/add_edit_epense_views/EditExpenseV.java
YinLingithub: resolve as not relevant

src/main/java/views/add_edit_epense_views/EditExpenseV.java
YinLingithub: I think the changes to EditExpenseV in this commit (fixed errors in edit use case) was unnecessary, as the code worked fine. 

The changes to EditCategoryV was the only thing needed to solve the problem in your code.

luluwuwu: my new commit in EditExpenseV is to fix another issue I found which is: When a user tries to edit an expense without selecting a category, the expense would automatically be assigned to the Other category since it is automatically selected. 

So to fix that problem, I decided to make sure the user inputs all required data before entering the use case. And what I did is to add the actionListener to categoryCombo and basically, it makes sure an action is performed. The conditions are added so that it is coordinated with EdotCategoryV's actionPerformed(). 

The issue in EditCategoryV that you mentioned(which is fixed!) is just a problem with the if-else statements such that certain ifs are not correctly accessed. (EditExpenseV had a similar problem in the same area but I also made changes to address the problem I mentioned above so it might be hard to tell)

YinLingithub: Though I don't think that would be a big problem. The point of having users to select a expense to edit is because the controller would take the input and add a new Expense if they don't. (similarly when in the EditCategoryV case, if a category is not selected, the controller will make a new Category). 

However, in this case, automatically selecting Other will not cause errors in the program: if they want to select Other, they do not have to touch the select Category, and the result will be what they want; if they want to select another Category, they will naturally click on the ComboBox. Moreover, the user can just edit the expense again if they make a mistake. 

The current implementation doesn't work properly, while the previous implementation (one before latest commit) works fine, so that's why I said to use the previous version. Of course, it's always better to have a implementation that does exactly what we want. However, as you will probably need to change some implementation in the UCI to implement/use the new Factory Design Pattern, I'd say the version you just had is already fulfilling the roles we need.

luluwuwu: I fixed that problem just now since it's just a problem with the if-else statements(again!! :O) in EditExpenseV. You will see it in my next commit!

luluwuwu: Just so you know this part that your suggesting is not the reason why its not working lol

YinLingithub (approved):
Nice fix! Everything works perfectly now!

aricasas (approved):
Approved fix!

katarinaavucic (approved):
Looks great!

PR #59, opened 2023-08-12
Address intellij warnings

### LINKED ISSUE #50: Address IntelliJ warnings

aricasas:
This PR will correct code warnings that IntelliJ displays for the project.


src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
```diff
@@ -52,11 +52,12 @@ private double toDouble(Object value) throws NumberFormatException, NullPointerE
...
      * @param categoryIDAdd CategoryID required for adding a new category to designated monthID MonthlyStorage Object.
      * @return CategoryOD String message indicating success/fail add attempt.
+     * @throws EntityException thrown when the new category input is invalid.
      */
     @Override
-    public CategoryOD addCategoryInMonth(CategoryID categoryIDAdd) {
+    public CategoryOD addCategoryInMonth(CategoryID categoryIDAdd) throws EntityException {
+        this.month = categoryIDAdd.getSession().getMonthlyData(categoryIDAdd.getMonthID());
         try {
-            this.month = categoryIDAdd.getSession().getMonthlyData(categoryIDAdd.getMonthID());
```
aricasas, line 59: @luluwuwu what do you think of this change. My reasoning can be seen in the commit message:

 Made addExpenseInMonth and addCategoryInMonth methods throw EntityException to fix the IntelliJ warning that only the interface was throwing but not any actual implementations. Also, I believe this is more correct since the EntityException caused by having the wrong monthID is caught in the view and shouldn't be caught here where the error message says it's because there's a category/expense with the same name.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -40,12 +40,13 @@ public static double toDouble(Object value) throws NumberFormatException {
...
+     * @throws EntityException thrown when the new expense input is invalid.
      */
     @Override
-    public ExpenseOD addExpenseInMonth(ExpenseID expenseIDAdd) {
+    public ExpenseOD addExpenseInMonth(ExpenseID expenseIDAdd) throws EntityException {
+        SessionStorage session = expenseIDAdd.getSession();
+        this.month = session.getMonthlyData(expenseIDAdd.getMonthID());
         try {
-            SessionStorage session = expenseIDAdd.getSession();
-            this.month = session.getMonthlyData(expenseIDAdd.getMonthID());
```
aricasas, line 48: @luluwuwu what do you think of this change. My reasoning can be seen in the commit message:

 Made addExpenseInMonth and addCategoryInMonth methods throw EntityException to fix the IntelliJ warning that only the interface was throwing but not any actual implementations. Also, I believe this is more correct since the EntityException caused by having the wrong monthID is caught in the view and shouldn't be caught here where the error message says it's because there's a category/expense with the same name.

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
katarinaavucic: I am not Lulu but this change makes sense to me!

src/main/java/entities/MonthObject.java
```diff
@@ -5,6 +5,7 @@
  * they can be created by MonthObjectFactory classes. It contains the setName and getName methods that all MonthObject
  * classes should have.
  */
+@SuppressWarnings("unused") // These methods are actually used, but just on the concrete class
```
katarinaavucic, line 8: What is the reasoning for having this here?

src/main/java/views/session_load/SessionLoadMenuVB.java
katarinaavucic, line 1: Why was this deleted? Will this cause an error when merging with main?

katarinaavucic (commented):
Everything looks good, just had a couple of questions!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: Right, the monthlyStorage.addExpense() and addCategory() methods throw the EntityException. 
But since currently this line is throwing the EntityException as well `this.month = session.getMonthlyData(expenseIDAdd.getMonthID());
`, should we move this line above the try catch so it doesn't mistake it?

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
YinLingithub: I think when Lulu and I talked about this, we said that although such an error should never occur in theory, it may cause the program to terminate if it's not caught anywhere in the code. Thus we decided to put this line of code inside the try catch. However, I do understand that this Exception does not match the warning message, so I'd say we better put it in another try catch that returns the correct warning.

luluwuwu: (Just moving this comment here)

Right, the monthlyStorage.addExpense() and addCategory() methods throw the EntityException.
But since currently this line is throwing the EntityException as well this.month = session.getMonthlyData(expenseIDAdd.getMonthID()); , should we move this line above the try catch so it doesn't mistake it?

luluwuwu: Or actually what Yin is saying is more correct

YinLingithub: Wait, never mind, I see that it gets caught in the view. So I agree is implementation is more correct!

YinLingithub: > Or actually what Yin is saying is more correct

Lulu, I think we both kind of forgot, but we had a conservation after that, where we said this is a different kind of exception so we want to catch it in the view.

luluwuwu: I believe before the view is able to catch that Exception, the catch within the use case is going to handle it no?

luluwuwu: That's why I suggest moving the line with getMonthlyData() outside the try-catch in ExpenseUCI. Let me know if that makes sense.

aricasas: Yep, moving the getMonthlyData outside the try catch is what I did

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
aricasas: Yep, moving the getMonthlyData outside the try catch is what I did

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
YinLingithub: > That's why I suggest moving the line with getMonthlyData() outside the try-catch in ExpenseUCI. Let me know if that makes sense.

I'm not sure I do. I believe it was inside the try catch previously, and now Ari is moving it outside, no? This change will result in the use case interactor throwing the issue directly to the view before it runs any code in the addCategoryInMonth().

src/main/java/views/session_load/SessionLoadMenuVB.java
aricasas: IntelliJ marked is as a warning because it was only implemented in one class, and it was only called inside of it. It isn't actually necessary to have this as an interface, and none of the other view classes use an interface for this. So I removed it to make it more consistent with everything else and to fix the warning.

src/main/java/entities/MonthObject.java
aricasas: It's to disregard an IntelliJ warning explicitly. Otherwise, IntelliJ gives a warning because we don't use the methods getName and setName anywhere with a MonthObject directly, but we do use them with the concrete classes, so it's not actually an unused method.

aricasas: Adding the decorator to the class makes IntelliJ not show that warning for the methods inside the class, and then I added a small comment explaining why it's ok to disregard this warning

src/main/java/views/session_load/SessionLoadMenuVB.java
aricasas: And it shouldn't cause any trouble when merging

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
aricasas: So does everyone think this change is all right?

src/main/java/entities/MonthObject.java
katarinaavucic: Makes sense, thank you!

src/main/java/views/session_load/SessionLoadMenuVB.java
katarinaavucic: Okay, makes sense thanks!

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
luluwuwu: I don't think that's going to happen? There isn't a try catch for this line so the code does not terminate after throwing the exception. (correct me if I'm wrong) 
I'm rather more concerned that if there is an EntityException thrown, the adding of an Expense is performed but it's an invalid month. However, I think this problem can be safely ignored since we know there shouldn't be an invalid monthID.

YinLingithub (commented):
I noticed that there's a warning for "Long surrounding method" in MonthMenuV's actionListener(). I wasn't sure what this meant when I was implementing this method, but I realized it is talking about code smells for having a long method. Although I do want everybody's opinion on this, as this warning never appeared in the createMonthView(), which is a even longer method.

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
katarinaavucic: Yes, I approve of this change.

katarinaavucic (approved):
Looks good!

katarinaavucic:
> I noticed that there's a warning for "Long surrounding method" in MonthMenuV's actionListener(). I wasn't sure what this meant when I was implementing this method, but I realized it is talking about code smells for having a long method. Although I do want everybody's opinion on this, as this warning never appeared in the createMonthView(), which is a even longer method.

I think if there was a warning in the current implementation Ari would have fixed it. I did not see anything pertaining to this in the commits so I think it might be alright.

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
YinLingithub: > I don't think that's going to happen? There isn't a try catch for this line so the code does not terminate after throwing the exception. (correct me if I'm wrong) 

As Ari pointe out, if you look at the AddExpenseV and EditExpenseV, there are try catch for catching is kind of exception.

luluwuwu: Yes, I was thinking it wrong. But I don't think that will be an issue either. 
Since if an EntityException is thrown due to an invalid monthID, we wouldn't want it to enter the use case. And if an EntityException is thrown due to a sameName error, it will be caught by the method within ExpenseUCI before it can be thrown in ExpenseV.

So I will approve this change.

luluwuwu (approved):
Nice thanks for taking care of all of these warnings!

YinLingithub (approved):
Looks good!

PR #60, opened 2023-08-12
46 implementation of factory design pattern in Add/Edit features

### LINKED ISSUE #46: Implementation of Factory Design Pattern

luluwuwu:
Implemented Add/Edit use cases with factory methods. Unable to run tests because main is not merged yet. Please ignore all errors and failed tests.

### DIFF
--- a/src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
+++ b/src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
@@ -13,7 +13,7 @@ import java.util.Objects;
  */
 public class CategoryUCI implements CategoryIB {
     private final CategoryOB categoryOB;
-    private final MonthObjectFactory categoryFactory;
+    private final CategoryFactory categoryFactory;
     private MonthlyStorage month;
 
 
@@ -62,7 +62,7 @@ public class CategoryUCI implements CategoryIB {
                 CategoryOD categoryODFailAdd = new CategoryOD("Category budget can't be less than $0. " +
                         "Please try again!");
                 return categoryOB.fail(categoryODFailAdd);}
-            month.addCategory(categoryFactory.createMonthObject(setCategoryCreatorInfo(categoryIDAdd)));
+            month.addCategory((Category) categoryFactory.createMonthObject(setCategoryCreatorInfo(categoryIDAdd)));
             CategoryOD categoryODSuccessAdd = new CategoryOD("You have added a new category!");
             return categoryOB.success(categoryODSuccessAdd);
 
@@ -144,14 +144,20 @@ public class CategoryUCI implements CategoryIB {
      * @return CategoryCreatorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory
      * for the createMonthObject method.
      */
-    private MonthObjectFactoryInputData setCategoryCreatorInfo(CategoryID categoryIDAdd){
-        return new CategoryCreatorInputData(categoryIDAdd.getName(), toDouble(categoryIDAdd.getValue()));}
+    private CategoryCreatorInputData setCategoryCreatorInfo(CategoryID categoryIDAdd){
+        CategoryCreatorInputData categoryCreatorInputData = new CategoryCreatorInputData();
+        categoryCreatorInputData.setName(categoryIDAdd.getName());
+        categoryCreatorInputData.setBudget(toDouble(categoryIDAdd.getValue()));
+        return categoryCreatorInputData;}
     /**
      * Sets information needed to edit a CategoryEditorInputData to call editMonthObject method in CategoryFactory
      * @return CategoryEditorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory
      * for the editMonthObject method.
      */
-    private MonthObjectFactoryInputData setCategoryEditorInfo(CategoryID categoryIDEdit){
-        return new CategoryEditorInputData(categoryIDEdit.getName(), toDouble(categoryIDEdit.getValue()),
-                findCategory(month.getCategoryData(), categoryIDEdit.getOldCategory()));}
+    private CategoryEditorInputData setCategoryEditorInfo(CategoryID categoryIDEdit){
+        CategoryEditorInputData categoryEditorInputData = new CategoryEditorInputData();
+        categoryEditorInputData.setName(categoryIDEdit.getName());
+        categoryEditorInputData.setBudget(toDouble(categoryIDEdit.getValue()));
+        categoryEditorInputData.setCategory(findCategory(month.getCategoryData(), categoryIDEdit.getOldCategory()));
+        return categoryEditorInputData;}
 }
--- a/src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
+++ b/src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
@@ -13,17 +13,12 @@ import java.util.Objects;
  */
 public class CategoryUCI implements CategoryIB {
     private final CategoryOB categoryOB;
-    private CategoryID categoryID;
-    private Category selectedCategory;
-    private double valueDouble;
-    private final CategoryFactory categoryFactory;
+    private final MonthObjectFactory categoryFactory;
     private MonthlyStorage month;
-    private SessionStorage session;
-    private int monthID;
 
 
     /**
-     * Constructs CategoryUCI.
+     * Constructs CategoryUCI. A CategoryFactory is constructed.
      * @param categoryP presenter that is related to the use case.
      */
     public CategoryUCI(CategoryOB categoryP) {
@@ -42,29 +37,32 @@ public class CategoryUCI implements CategoryIB {
 
     /**
      * Overrides method in CategoryIB.
-     * Attempts to add a category with information from CategoryID and returns a CategoryOD indicating whether fail/success after execution.
+     * Attempts to add a category with information from CategoryID and returns a CategoryOD
+     * indicating whether fail/success after execution.
+     * CategoryFactory methods are implemented to better adhere to Liskov's Substitution Principle.
      * Provides detailed fail messages according to each condition below. (Explained in comments.)
      * NOTE: There are actually two different EntityException thrown.
      * One is from the SessionStorage Object when checking if monthID is in session.
-     *  (Although we know MonthlyStorage with monthID is always in the SessionStorage, it will be caught at views/add_edit_category_views/AddCategoryV.java).
-     * The second one is from the creation of a Category Object implementing addCategory() from MonthlyStorage, and is caught in the current implementation!
+     *  (Although we know MonthlyStorage with monthID is always in the SessionStorage,
+     *  it will be caught at views/add_edit_category_views/AddCategoryV.java).
+     * The second one is from the creation of a Category Object implementing addCategory() from MonthlyStorage,
+     * and is caught in the current implementation!
      *
-     * @param categoryIDAdd CategoryID required for adding a new category to the designated monthID MonthlyStorage Object.
+     * @param categoryIDAdd CategoryID required for adding a new category to designated monthID MonthlyStorage Object.
      * @return CategoryOD String message indicating success/fail add attempt.
      */
     @Override
     public CategoryOD addCategoryInMonth(CategoryID categoryIDAdd){
-        try{this.categoryID = categoryIDAdd;
-            this.session = categoryID.getSession();
-            this.monthID = categoryID.getMonthID();
-            this.month = session.getMonthlyData(monthID);
-            this.valueDouble = toDouble(categoryID.getValue());
+        try{
+            this.month = categoryIDAdd.getSession().getMonthlyData(categoryIDAdd.getMonthID());
+            double valueDouble = toDouble(categoryIDAdd.getValue());
 
             if (valueDouble < 0) {
                 // Category budget less than 0: User tries to add a new budget value that is a negative number.
-                CategoryOD categoryODFailAdd = new CategoryOD("Category budget can't be less than $0. Please try again!");
+                CategoryOD categoryODFailAdd = new CategoryOD("Category budget can't be less than $0. " +
+                        "Please try again!");
                 return categoryOB.fail(categoryODFailAdd);}
-            month.addCategory(categoryFactory.createMonthObject(setCategoryCreatorInfo()));
+            month.addCategory(categoryFactory.createMonthObject(setCategoryCreatorInfo(categoryIDAdd)));
             CategoryOD categoryODSuccessAdd = new CategoryOD("You have added a new category!");
             return categoryOB.success(categoryODSuccessAdd);
 
@@ -74,15 +72,18 @@ public class CategoryUCI implements CategoryIB {
             return categoryOB.fail(categoryODFailAdd);
         } catch (EntityException e) {
             //EntityException: User tries to add an invalid Category name but failed. (See entities/EntityException.java)
-            CategoryOD categoryODFailAdd = new CategoryOD("There is already a category with this new name in this month.");
+            CategoryOD categoryODFailAdd = new CategoryOD("There is already a category with this new name " +
+                    "in this month.");
             return categoryOB.fail(categoryODFailAdd);
         }
     }
     /**
      * Overrides method in CategoryIB.
-     * Attempts to edit a category with information from CategoryID and returns a CategoryOD indicating whether fail/success after execution.
+     * Attempts to edit a category with information from CategoryID and returns a CategoryOD
+     * indicating whether fail/success after execution.
+     * CategoryFactory methods are implemented to better adhere to Liskov's Substitution Principle.
      * Provides detailed fail messages according to each condition below. (Explained in comments.)
-     * @param categoryIDEdit CategoryID required for editing a new category to the designated monthID MonthlyStorage Object.
+     * @param categoryIDEdit CategoryID required for editing a new category to designated monthID MonthlyStorage Object.
      * @return CategoryOD String message indicating success/fail add attempt.
      * @throws EntityException (Although we know MonthlyStorage with monthID is always in the SessionStorage,
      *                          it will be caught at views/add_edit_category_views/EditCategoryV.java).
@@ -90,37 +91,37 @@ public class CategoryUCI implements CategoryIB {
     @Override
     public CategoryOD editCategoryInMonth(CategoryID categoryIDEdit) throws EntityException{
 
-        try {this.categoryID = categoryIDEdit;
-            this.session = categoryID.getSession();
-            this.monthID = categoryID.getMonthID();
-            this.month = session.getMonthlyData(monthID);
+        try {
+            this.month = categoryIDEdit.getSession().getMonthlyData(categoryIDEdit.getMonthID());
             ArrayList<Category> monthCategoryList = month.getCategoryData();
-            this.valueDouble = toDouble(categoryID.getValue());
+            double valueDouble = toDouble(categoryIDEdit.getValue());
 
             if (valueDouble < 0) {
                 //Category budget less than 0: User tries to edit a budget value with input that is a negative number.
-                CategoryOD categoryODFailEdit = new CategoryOD("Category budget can't be less than $0. Please try again!");
+                CategoryOD categoryODFailEdit = new CategoryOD("Category budget can't be less than $0. " +
+                        "Please try again!");
                 return categoryOB.fail(categoryODFailEdit);}
 
-            this.selectedCategory = findCategory(monthCategoryList, categoryIDEdit.getOldCategory());
-
-            if(!Objects.equals(categoryID.getName(), categoryIDEdit.getOldCategory())){
+            if(!Objects.equals(categoryIDEdit.getName(), categoryIDEdit.getOldCategory())){
                 for (Category category1 : monthCategoryList) {
-                if (category1.getName().equals(categoryID.getName())) {
+                if (category1.getName().equals(categoryIDEdit.getName())) {
                     //Repeated name: User tries to edit category name to another name that exists in the month.
-                    CategoryOD categoryODFailEdit = new CategoryOD("There is already a category with this new name in this month.");
+                    CategoryOD categoryODFailEdit = new CategoryOD("There is already a category with this new name " +
+                            "in this month.");
                     return categoryOB.fail(categoryODFailEdit);}}}
 
-            categoryFactory.editMonthObject(setCategoryEditorInfo());
+            categoryFactory.editMonthObject(setCategoryEditorInfo(categoryIDEdit));
             CategoryOD categoryODSuccessEdit = new CategoryOD("You have edited a category!");
             return categoryOB.success(categoryODSuccessEdit);
 
         } catch (NoSuchElementException e) {
             //NoSuchElementException: User tries to edit a category that does not exist.
-            CategoryOD categoryODFailEdit = new CategoryOD("There is no such category in the current month. Please add a new category or select existing category!");
+            CategoryOD categoryODFailEdit = new CategoryOD("There is no such category in the current month. " +
+                    "Please add a new category or select existing category!");
             return categoryOB.fail(categoryODFailEdit);
         } catch(NumberFormatException|NullPointerException e){
-            //NumberFormatException|NullPointerException: User tries to edit a budget value with input that can not be converted to a double.
+            //NumberFormatException|NullPointerException: User tries to edit a budget value
+            // with input that can not be converted to a double.
             CategoryOD categoryODFailEdit = new CategoryOD("Category budget needs to be a number. Please try again!");
             return categoryOB.fail(categoryODFailEdit);
         }
@@ -139,15 +140,18 @@ public class CategoryUCI implements CategoryIB {
                 return c;}}
         throw new NoSuchElementException();}
     /**
-     * Sets the information needed to create a CategoryCreatorInputData to call the createMonthObject method in CategoryFactory
-     * @return CategoryCreatorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory for the createMonthObject method.
+     * Sets information needed to create a CategoryCreatorInputData to call createMonthObject method in CategoryFactory
+     * @return CategoryCreatorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory
+     * for the createMonthObject method.
      */
-    private CategoryCreatorInputData setCategoryCreatorInfo(){
-        return CategoryCreatorInputData categoryCreatorInputData = new CategoryCreatorInputData(categoryID.getName(), valueDouble);}
+    private MonthObjectFactoryInputData setCategoryCreatorInfo(CategoryID categoryIDAdd){
+        return new CategoryCreatorInputData(categoryIDAdd.getName(), toDouble(categoryIDAdd.getValue()));}
     /**
-     * Sets the information needed to edit a CategoryEditorInputData to call the editMonthObject method in CategoryFactory
-     * @return CategoryEditorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory for the editMonthObject method.
+     * Sets information needed to edit a CategoryEditorInputData to call editMonthObject method in CategoryFactory
+     * @return CategoryEditorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory
+     * for the editMonthObject method.
      */
-    private CategoryEditorInputData setCategoryEditorInfo(){
-        return CategoryEditorInputData categoryEditorInputData(categoryID.getName(), valueDouble, selectedCategory);}
+    private MonthObjectFactoryInputData setCategoryEditorInfo(CategoryID categoryIDEdit){
+        return new CategoryEditorInputData(categoryIDEdit.getName(), toDouble(categoryIDEdit.getValue()),
+                findCategory(month.getCategoryData(), categoryIDEdit.getOldCategory()));}
 }
--- a/src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
+++ b/src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
@@ -13,7 +13,7 @@ import java.util.Objects;
  */
 public class CategoryUCI implements CategoryIB {
     private final CategoryOB categoryOB;
-    private final CategoryFactory categoryFactory;
+    private final MonthObjectFactory categoryFactory;
     private MonthlyStorage month;
 
 
@@ -62,7 +62,8 @@ public class CategoryUCI implements CategoryIB {
                 CategoryOD categoryODFailAdd = new CategoryOD("Category budget can't be less than $0. " +
                         "Please try again!");
                 return categoryOB.fail(categoryODFailAdd);}
-            month.addCategory((Category) categoryFactory.createMonthObject(setCategoryCreatorInfo(categoryIDAdd)));
+            Category newCategory = (Category) categoryFactory.createMonthObject(setCategoryCreatorInfo(categoryIDAdd));
+            month.addCategory(newCategory);
             CategoryOD categoryODSuccessAdd = new CategoryOD("You have added a new category!");
             return categoryOB.success(categoryODSuccessAdd);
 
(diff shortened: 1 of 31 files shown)


src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
```diff
@@ -14,10 +11,16 @@
...
+    private Category selectedCategory;
+    private double valueDouble;
     private final CategoryFactory categoryFactory;
+    private MonthlyStorage month;
+    private SessionStorage session;
+    private int monthID;
+
 
     /**
      * Constructs CategoryUCI.
```
katarinaavucic: Nice construction of the CategoryFactory!

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
```diff
@@ -14,10 +11,16 @@
  * Implements CategoryIB.
  * Creates CategoryOD objects related to specific use case fail/success conditions.
  */
-
 public class CategoryUCI implements CategoryIB {
     private final CategoryOB categoryOB;
+    private CategoryID categoryID;
+    private Category selectedCategory;
+    private double valueDouble;
     private final CategoryFactory categoryFactory;
```
katarinaavucic: I think you should store it as "private final MonthObjectFactory categoryFactory" instead, to demonstrate Liskov's Substitution Principle.

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
```diff
@@ -131,4 +138,16 @@ public Category findCategory(ArrayList<Category> monthCategoryData, String name)
             if (Objects.equals(c.getName(), name)){
                 return c;}}
         throw new NoSuchElementException();}
+    /**
+     * Sets the information needed to create a CategoryCreatorInputData to call the createMonthObject method in CategoryFactory
+     * @return CategoryCreatorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory for the createMonthObject method.
+     */
+    private CategoryCreatorInputData setCategoryCreatorInfo(){
```
katarinaavucic: These two private methods should return type MonthObjectFactoryInputData instead! This would better demonstrate Liskov's Substitution Principle.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -9,8 +9,16 @@
 public class ExpenseUCI implements ExpenseIB {
     private final ExpenseOB expenseOB;
     private ExpenseID expenseID;
+    private Category selectedCategory;
+    private Expense selectedExpense;
+    private double valueDouble;
+    private ArrayList<Expense> monthExpenseList;
+    private ArrayList<Category> monthCategoryList;
     private ArrayList<Expense> recurringExpenseList;
     private final ExpenseFactory expenseFactory;
```
katarinaavucic: Like above, storing as MonthObjectFactory instead.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -201,17 +210,17 @@ private boolean changeInRecurringInfo() {
...
     /**
-     * Edits selected Expense prepared with necessary edit info from expenseID
-     * @param expenseID relevant expenseID
-     * @param valueDouble new expense value for edit
-     * @param selectedExpense Expense wish to edit
-     * @param selectedCategory Category the selected Expense belongs to
+     * Sets the information needed to create an ExpenseCreatorInputData to call the createMonthObject method in ExpenseFactory
+     * @return ExpenseCreatorInputData MonthObjectFactoryInputData Object specifically used in ExpenseFactory for the createMonthObject method.
+     */
+    private ExpenseCreatorInputData setExpenseCreatorInfo(){
```
katarinaavucic: Like above as well, return MonthObjectFactoryInputData.

katarinaavucic: For setExpenseEditorInfo too.

katarinaavucic: When you do these dont forget to update all four Javadocs!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -9,8 +9,16 @@
 public class ExpenseUCI implements ExpenseIB {
     private final ExpenseOB expenseOB;
```
katarinaavucic, line 10: I noticed in both CategoryUCI and ExpenseUCI there are a lot of attributes saved. I think you should look at how many uses these have and get rid of the ones that are only used in 1 or 2 methods! I think the reason many of them are here is because you use them in private methods, but I recommend passing them in instead. Anyone who uses our code and tries to understand how the method works would likely get confused by all of these attributes.

katarinaavucic: Make sure you keep the Factory and Presenter and anything used in both, but some attributes like name and value are likely not necessary to store here!

katarinaavucic (commented):
YAYYYY its coming together!! All of the usages look good, I just had a few comments on casting as well as unnecessary attributes to clean up the code!

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
luluwuwu: I don't really mind, but wouldn't it be a little confusing since this use case is only ever going to be dealing with Category entity?

aricasas: Yes but we want the code to be as general as possible. And since the only method you're gonna be using from the factory is a method contained in the abstract class, you can just keep the type signature for the factory abstract too

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
```diff
@@ -139,15 +140,18 @@ public Category findCategory(ArrayList<Category> monthCategoryData, String name)
...
-     * Sets the information needed to create a CategoryCreatorInputData to call the createMonthObject method in CategoryFactory
-     * @return CategoryCreatorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory for the createMonthObject method.
+     * Sets information needed to create a CategoryCreatorInputData to call createMonthObject method in CategoryFactory
+     * @return CategoryCreatorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory
+     * for the createMonthObject method.
      */
-    private CategoryCreatorInputData setCategoryCreatorInfo(){
-        return CategoryCreatorInputData categoryCreatorInputData = new CategoryCreatorInputData(categoryID.getName(), valueDouble);}
+    private MonthObjectFactoryInputData setCategoryCreatorInfo(CategoryID categoryIDAdd){
+        return new CategoryCreatorInputData(categoryIDAdd.getName(), toDouble(categoryIDAdd.getValue()));}
```
katarinaavucic: You need to change these to accommodate for the setters instead of the constructor (so creating the object and then calling setName and setBudget).

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
katarinaavucic: Why did you switch it back to CategoryFactory?

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
```diff
@@ -144,14 +144,20 @@ public Category findCategory(ArrayList<Category> monthCategoryData, String name)
      * @return CategoryCreatorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory
      * for the createMonthObject method.
      */
-    private MonthObjectFactoryInputData setCategoryCreatorInfo(CategoryID categoryIDAdd){
-        return new CategoryCreatorInputData(categoryIDAdd.getName(), toDouble(categoryIDAdd.getValue()));}
+    private CategoryCreatorInputData setCategoryCreatorInfo(CategoryID categoryIDAdd){
```
katarinaavucic: This should be cast to MonthObjectFactoryInputData as previously discussed.

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
```diff
@@ -144,14 +144,20 @@ public Category findCategory(ArrayList<Category> monthCategoryData, String name)
...
+        return categoryCreatorInputData;}
     /**
      * Sets information needed to edit a CategoryEditorInputData to call editMonthObject method in CategoryFactory
      * @return CategoryEditorInputData MonthObjectFactoryInputData Object specifically used in CategoryFactory
      * for the editMonthObject method.
      */
-    private MonthObjectFactoryInputData setCategoryEditorInfo(CategoryID categoryIDEdit){
-        return new CategoryEditorInputData(categoryIDEdit.getName(), toDouble(categoryIDEdit.getValue()),
-                findCategory(month.getCategoryData(), categoryIDEdit.getOldCategory()));}
+    private CategoryEditorInputData setCategoryEditorInfo(CategoryID categoryIDEdit){
```
katarinaavucic: Same here.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: This one you made ExpenseFactory again?

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -63,13 +60,13 @@ public ExpenseOD addExpenseInMonth(ExpenseID expenseIDAdd){
...
             if (expenseIDAdd.getIsRecurringExpense()) {
-                Expense newrecurExpense = expenseFactory.createMonthObject(setExpenseCreatorInfo(expenseIDAdd));
+                Expense newrecurExpense = (Expense) expenseFactory.createMonthObject(setExpenseCreatorInfo(expenseIDAdd));
                 session.addRecurExpense(newrecurExpense);
                 month.addExpense(newrecurExpense);
                 ExpenseOD expenseODSuccessAdd = new ExpenseOD("You have created a new recurring expense!");
                 return expenseOB.success(expenseODSuccessAdd);
             }else{
-                month.addExpense(expenseFactory.createMonthObject(setExpenseCreatorInfo(expenseIDAdd)));
+                month.addExpense((Expense) expenseFactory.createMonthObject(setExpenseCreatorInfo(expenseIDAdd)));
```
katarinaavucic: I think every time you call the factory you should make a new variable and cast it to Expense and then add it to the month (like you did with newRecurExpense. I think this would make the code more readable than three layered method calls.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -63,13 +60,13 @@ public ExpenseOD addExpenseInMonth(ExpenseID expenseIDAdd){
                     return expenseOB.fail(expenseODFailEdit);}}
 
             if (expenseIDAdd.getIsRecurringExpense()) {
-                Expense newrecurExpense = expenseFactory.createMonthObject(setExpenseCreatorInfo(expenseIDAdd));
+                Expense newrecurExpense = (Expense) expenseFactory.createMonthObject(setExpenseCreatorInfo(expenseIDAdd));
```
katarinaavucic: Should be newRecurExpense (camel case).

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -120,14 +112,14 @@ public ExpenseOD editExpenseInMonth(ExpenseID expenseIDEdit) throws EntityExcept
...
+                    session.addRecurExpense((Expense) expenseFactory.editMonthObject(setExpenseEditorInfo(expenseIDEdit)));
                     // Success edit to new recurring expense.
                     ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes " +
-                            "of this new recurring expense!");
+                            "of this recurring expense!");
                     return expenseOB.success(expenseODSuccessEditRecurring);
                 }else{
                     expenseFactory.editMonthObject(setExpenseEditorInfo(expenseIDEdit));
-                    session.deleteRecurExpense(expenseIDEdit.getName());
+                    session.deleteRecurExpense(expenseIDEdit.getOldExpense());
```
katarinaavucic: Why are you using the old name? When the Expense is edited to have the new name that should change in the recurring expense list so you should be using getName() like before.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -157,7 +149,7 @@ public ExpenseOD editExpenseInMonth(ExpenseID expenseIDEdit) throws EntityExcept
      * @throws NoSuchElementException thrown when couldn't find Expense with String name.
      */
         public Expense findExpense(ArrayList<Expense> expenseData, String name)throws NoSuchElementException{
-            if(expenseData != null) {
+            if(!expenseData.isEmpty()) {
```
katarinaavucic: Why did you make this switch? If the program runs null.isEmpty() it will throw a Null Pointer Exception.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -208,18 +200,24 @@ private boolean changeInRecurringInfo(ExpenseID expenseID) {
      * @return ExpenseCreatorInputData MonthObjectFactoryInputData Object specifically used in ExpenseFactory
      * for the createMonthObject method.
      */
-    private MonthObjectFactoryInputData setExpenseCreatorInfo(ExpenseID expenseIDAdd){
-        return new ExpenseCreatorInputData(expenseIDAdd.getName(), findCategory(month.getCategoryData(),
-                expenseIDAdd.getCategory()), toDouble(expenseIDAdd.getValue()));}
+    private ExpenseCreatorInputData setExpenseCreatorInfo(ExpenseID expenseIDAdd){
```
katarinaavucic: Should be cast to MonthObjectFactoryInputData.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -208,18 +200,24 @@ private boolean changeInRecurringInfo(ExpenseID expenseID) {
...
     /**
      * Sets the information needed to edit an ExpenseEditorInputData to call editMonthObject method in ExpenseFactory
      * @return ExpenseEditorInputData MonthObjectFactoryInputData Object specifically used in ExpenseFactory
      * for the editMonthObject method.
      */
-    private MonthObjectFactoryInputData setExpenseEditorInfo(ExpenseID expenseIDEdit){
-        return new ExpenseEditorInputData(expenseIDEdit.getName(),findCategory(month.getCategoryData(),
-                expenseIDEdit.getCategory()), toDouble(expenseIDEdit.getValue()),
-                findExpense(month.getExpenseData(), expenseIDEdit.getOldExpense()));}
+    private ExpenseEditorInputData setExpenseEditorInfo(ExpenseID expenseIDEdit){
```
katarinaavucic: Cast here as well.

katarinaavucic (changes requested):
I think your merge to main might have reset your previous commit with the cast changes! Also had some questions regarding why you changed some things. Other than that the implementation looks good!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
YinLingithub, line 1: I think the fix you made for the recurring expense case in ExpenseUCI and EditExpenseV are overwritten by this branch. Make sure you copy those code in here.

src/main/java/use_cases/add_edit_category_use_case/CategoryUCI.java
luluwuwu: MonthObjectFactory isn't public so I'm not sure how I can declare it.

YinLingithub: > MonthObjectFactory isn't public so I'm not sure how I can declare it.

I believe the MonthObjectFactory is a public class, so you could just declare it like how you declare any other class object. From what I understand, it should work somewhat like this:

MonthObjectFactoryInputData factoryInput = new CategoryCreaterInputData(input goes here);
this.categoryFactory = new MonthObjectFactory(factoryInput);

katarinaavucic: They are actually abstract so they are not public! I apologize, it looks like Ari fixed this in his Intellij warning branch. https://github.com/CSC207-2023Y-UofT/course-project-DoorKeepers/pull/59/commits/90719a0e91d12b4fc309817448c2a4d340006a88

katarinaavucic: You could just add the public modifiers in your version to make sure it works.

luluwuwu: I've made them public but that still doesn't change the fact that I can't instantiate an abstract class. 
So: `this.categoryFactory = new `MonthObjectFactory(factoryInput);` won't work?

YinLingithub: > I've made them public but that still doesn't change the fact that I can't instantiate an abstract class. So: `this.categoryFactory = new `MonthObjectFactory(factoryInput);` won't work?

Ah, my bad. You should declare the private attribute like this:
private final MonthObjectFactory categoryFactory;

then do:
this.categoryFactory = new CategoryFactory(factoryInput);

luluwuwu: Screen Shot 2023-08-14 at 5 29 04 PM
This is what happens when I try casting MonthObjectFactory.

aricasas: You can declare it as the CategoryCreatorInputData and only cast it at the return statement

katarinaavucic: > I've made them public but that still doesn't change the fact that I can't instantiate an abstract class.
> So: `this.categoryFactory = new `MonthObjectFactory(factoryInput);` won't work?

Also just to mention here, the reason why you can’t make an instance of it is because it is abstract, and java prevents you from instantiating an abstract class! So casting in the return like Ari said is the way you are supposed to do it.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: In this case, the condition describes when changeinRecurringInfo is true and the new input is not a recurring expense. This means deleting the recurring expense should be called upon the old expense(to remove the selected expense from recurring list). expenseIDEdit.getName is the new name that is given to the oldExpense, and won't be recorded in the recurring list.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: Yup that is a mistake tks!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: I believe they're all up to date, which parts are you talking about?

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: Hmm, but the recurring list in the session is an ArrayList of Expense objects, meaning that changing the Expense objects name to a new name would update in that list (especially since it’s not a local copy you made, it’s the one stored in the session). So then the recurring expense list SHOULD hold the Expense with the new name and not the old one (otherwise that means that our code does not work, as our Expense objects are not updating).

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: No worries!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: Remember this is only the case when the user is changing the isRecurring from True to False, hence the behaviour should be to remove the selected/old expense from the recurring list and set other changes. What you described is the general case of editing a name (and no change in isRecurring data), which is handled by line125!

YinLingithub: Kat, it because Lulu is doing this before she edits the expense, so she needs to access the Expense object using the old name.

luluwuwu: Not quite actually lol

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
YinLingithub: I think you've solve it somehow in more current commits.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -181,7 +181,7 @@ private boolean checkHaveSameNameInList(ArrayList<Expense> expenseList, String n
      * (The user changes their response for the isRecurringExpense checkbox)
      * @return boolean suggesting there is a need to update Session.recurringExpenseData
      */
-    private boolean changeInRecurringInfo() {
+    private boolean changeInRecurringInfo(ExpenseID expenseID) {
```
aricasas, line 220: While trying to review I kept getting confused of what this method was doing, but after looking at it I think it could be written in a clearer way, especially its JavaDoc. I changed it as an example

```java
    /**
     * Returns whether there's a difference in the recurring status of the old expense and the edited expense.
     * This could mean that the expense wasn't recurring and now the user is trying to make it recurring,
     * or vice versa.
     *
     * @return a boolean describing whether there was a change in the recurring status of the expense
     */
    private boolean changeInRecurringInfo(ExpenseID expenseID) {
        boolean isOldExpenseRecurring;
        try {
            findExpense(expenseID.getSession().getRecurData(), expenseID.getOldExpense());
            isOldExpenseRecurring = false;
        } catch (NoSuchElementException e) {
            isOldExpenseRecurring = true;
        }

        return isOldExpenseRecurring != expenseID.getIsRecurringExpense();
    }
```

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -190,22 +190,31 @@ private boolean changeInRecurringInfo() {
...
-    private void setEditInfo(ExpenseID expenseID, double valueDouble, Expense selectedExpense, Category selectedCategory){
-        selectedExpense.setName(expenseID.getName());
-        expenseID.setValue(valueDouble);
-        selectedExpense.setValue(valueDouble);
-        selectedExpense.setCategory(selectedCategory);}
-
-}
-
-
+    private MonthObjectFactoryInputData setExpenseCreatorInfo(ExpenseID expenseIDAdd){
```
aricasas: Since this method is making a Factory Input Data, why not use the words Input Data instead of info for the method name. I think that could be clearer

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -190,22 +190,31 @@ private boolean changeInRecurringInfo() {
...
+        expenseCreatorInputData.setValue(toDouble(expenseIDAdd.getValue()));
+        expenseCreatorInputData.setCategory(findCategory(month.getCategoryData(), expenseIDAdd.getCategory()));
+        return expenseCreatorInputData;}
 
+    /**
+     * Sets the information needed to edit an ExpenseEditorInputData to call editMonthObject method in ExpenseFactory
+     * @return ExpenseEditorInputData MonthObjectFactoryInputData Object specifically used in ExpenseFactory
+     * for the editMonthObject method.
+     */
+    private MonthObjectFactoryInputData setExpenseEditorInfo(ExpenseID expenseIDEdit){
```
aricasas: Same for this one. Why not Input Data instead of info

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -30,44 +32,42 @@ public static double toDouble(Object value) throws NumberFormatException {
...
+
             if (valueDouble < 0) {
                 //Expense value less than 0 fail: When a user tries to add the expense value with a negative number.
                 ExpenseOD expenseODFailAdd = new ExpenseOD("Expense value can't be less than $0. Please try again!");
                 return expenseOB.fail(expenseODFailAdd);}
 
-            for (Expense expense1 : monthExpenseList) {
-                if (expense1.getName().equals(expenseID.getName())) {
+            for (Expense expense1 : month.getExpenseData()) {
+                if (expense1.getName().equals(expenseIDAdd.getName())) {
```
aricasas: Why aren't you using your helper method checkHaveSameNameInList for checking this?

aricasas: Also, I just realized that the helper method is buggy. Inside its for loop it returns the first boolean that it checks without actually going through the whole list.

aricasas: Here's a quick suggestion on how to fix it
```java
    private boolean checkHaveSameNameInList(ArrayList<Expense> expenseList, String name) {
        for (Expense expense1 : expenseList) {
            if (expense1.getName().equals(name)) {
                return true;
            }
        }
        return false;
    }
```

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -88,49 +89,48 @@ public ExpenseOD addExpenseInMonth(ExpenseID expenseIDAdd) throws EntityExceptio
...
-                    session.addRecurExpense(selectedExpense);
+
+            if (changeInRecurringInfo(expenseIDEdit)) {
+                if (expenseIDEdit.getIsRecurringExpense()) {
+                    Expense newRecurExpense = (Expense) expenseFactory.editMonthObject(setExpenseEditorInfo(expenseIDEdit));
+                    session.addRecurExpense(newRecurExpense);
                     // Success edit to new recurring expense.
-                    ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes of this recurring expense!");
+                    ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes " +
+                            "of this recurring expense!");
```
aricasas: Maybe you could add a comment explaining what exactly is happening here so it's more understandable

Something like
```java
            if (changeInRecurringInfo(expenseIDEdit)) {
                if (expenseIDEdit.getIsRecurringExpense()) {
                    // If expense is newly recurring, we have to update it inside MonthlyStorage (where it currently is)
                    // and also add it to the recurring expenses list in SessionStorage
                    Expense newRecurExpense = (Expense) expenseFactory.editMonthObject(setExpenseEditorInfo(expenseIDEdit));
                    session.addRecurExpense(newRecurExpense);
                    // Success edit to new recurring expense.
                    ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes " + "of this recurring expense!");
                    return expenseOB.success(expenseODSuccessEditRecurring);
  
```

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
katarinaavucic: We decided that getName() is correct on call.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
```diff
@@ -88,49 +89,48 @@ public ExpenseOD addExpenseInMonth(ExpenseID expenseIDAdd) throws EntityExceptio
...
-                    setEditInfo(expenseID, valueDouble, selectedExpense,selectedCategory);
-                    session.deleteRecurExpense(selectedExpense.getName());
+                    expenseFactory.editMonthObject(setExpenseEditorInfo(expenseIDEdit));
+                    session.deleteRecurExpense(expenseIDEdit.getName());
                     //Success edit to remove recurring expense.
-                    ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes of " +
-                            "this expense and it is no longer a recurring expense!");
+                    ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes " +
+                            "of this expense and it is no longer a recurring expense!");
                     return expenseOB.success(expenseODSuccessEditRecurring);}
```
aricasas: Maybe add a comment here too. Something like

```java
                } else {
                    // If expense was recurring but not anymore, alongside editing it we also have to remove it from
                    // the recurring expenses list
                    expenseFactory.editMonthObject(setExpenseEditorInfo(expenseIDEdit));
                    session.deleteRecurExpense(expenseIDEdit.getName());
                    //Success edit to remove recurring expense.
                    ExpenseOD expenseODSuccessEditRecurring = new ExpenseOD("You have updated all changes " + "of this expense and it is no longer a recurring expense!");
                    return expenseOB.success(expenseODSuccessEditRecurring);
                }
```

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
aricasas, line 1: Also, I think if you make IntelliJ format your file it could be a lot more readable. Right now there are a bunch of nested statements, but since your brackets are a bit all over the place it makes it harder to understand where scopes begin and end. I think <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>L</kbd> makes it autoformat the file? Or at least that's the keyboard shortcut on my computer, just press it every once in a while so IntelliJ takes care of the formatting for you

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
aricasas: Or Factory Input Data if you don't want it to be confused for the UCI input data. Something likesetExpenseCreatorFactoryInputData. Yes the name is very long but to be honest I think that would be clearer for someone reading the code for the first time. I had to keep shuffling up and down while doing my review since I didn't understand just from the name what these methods did

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
aricasas: Maybe you can change the method name to isExpenseNameInList. A bit shorter and I think it could be clearer on what it does

aricasas (changes requested):
I found a bug in a helper method in ExpenseUCI. Other than that everything seems correct. I like that you're using many helper methods to split up the code, but I did have some comments on them to improve their clarity, because their names kept confusing me while I was doing the review.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: I admit it is a little confusing, and I never thought of assigning a variable. If you don't mind, I will change it to your version!

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: Oh I did not realise, this is such a bug, thanks Ari!
And I will fix the name as well.

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: I think the logic in your implementation is flipped. I believe the following will be correct!
```
 private boolean changeInRecurringInfo(ExpenseID expenseID) {
        boolean isOldExpenseRecurring;
        try {
            findExpense(expenseID.getSession().getRecurData(), expenseID.getOldExpense());
            isOldExpenseRecurring = true;
        } catch (NoSuchElementException e) {
            isOldExpenseRecurring = false;
        }
        return isOldExpenseRecurring != expenseID.getIsRecurringExpense();
    }
```

src/main/java/use_cases/add_edit_expenses_use_case/ExpenseUCI.java
luluwuwu: This is brilliant !

YinLingithub (approved):
Everything looks good! Yay, well done!!

katarinaavucic (approved):
Perfect! I can definitely read the code more clearly now, and everything looks great functionality-wise!

aricasas (approved):
Everything looks great!

